PR #8, opened 2022-12-13
raft: support asynchronous storage writes

nvanbenschoten:
Fixes https://github.com/etcd-io/etcd/issues/12257.

**NOTE: this PR was originally presented in https://github.com/etcd-io/etcd/pull/14627 before `etcd/raft` was extracted into a separate repository (see https://github.com/etcd-io/etcd/issues/14713). Review comments from the original PR have been addressed.**

This change adds opt-in support to raft to **perform local storage writes asynchronously** from the raft state machine handling loop.

## Summary

A new `AsyncStorageWrites` configuration instructs the raft node to write to its local storage (raft log and state machine) using a request/response message passing interface instead of the default `Ready`/`Advance` function call interface. Local storage messages can be pipelined and processed asynchronously (with respect to `Ready` iteration), facilitating reduced interference between Raft proposals and increased batching of log appends and state machine application. As a result, use of asynchronous storage writes can reduce end-to-end commit latency and increase maximum throughput.

When `AsyncStorageWrites` is enabled, the `Ready.Message` slice will include new `MsgStorageAppend` and `MsgStorageApply` messages. The messages will target a `LocalAppendThread` and a `LocalApplyThread`, respectively. Messages to the same target must be reliably processed in order. In other words, they can't be dropped (like messages over the network) and those targeted at the same thread can't be reordered. Messages to different targets can be processed in any order.

`MsgStorageAppend` carries Raft log entries to append, election votes to persist, and snapshots to apply. All writes performed in response to a `MsgStorageAppend` are expected to be durable. The message assumes the role of the Entries, HardState, and Snapshot fields in Ready.

`MsgStorageApply` carries committed entries to apply. The message assumes the role of the CommittedEntries field in Ready.

Local messages each carry one or more response messages which should be delivered after the corresponding storage write has been completed. These responses may target the same node or may target other nodes. The storage threads are not responsible for understanding the response messages, only for delivering them to the correct target after performing the storage write.

## Design Considerations

- There must be no regression for existing users that do not enable `AsyncStorageWrites`. For instance, CommittedEntries must not wait on unstable entries to be stabilized in cases where a follower is given committed entries in a MsgApp.
- Asynchronous storage work should use a message passing interface, like the rest of this library.
- The Raft leader and followers should behave symmetrically. Both should be able to use asynchronous storage writes for log appends and entry application.
- The LocalAppendThread on a follower should be able to send MsgAppResp messages directly to the leader without passing back through the raft state machine handling loop.
- The `unstable` log should remain true to its name. It should hold entries until they are stable and should not rely on an intermediate reliable cache.
- Pseudo-targets should be assigned to messages that target the local storage systems to denote required ordering guarantees.
- Code should be maximally unified across `AsyncStorageWrites=false` and `AsyncStorageWrites=true`. `AsyncStorageWrites=false` should be a special case of `AsyncStorageWrites=true` where the library hides the possibility of asynchrony.
- It should be possible to apply snapshots asynchronously, even though a snapshot touches both the Raft log state and the state machine. The library should make this easy for users to handle by delaying all committed entries until after the snapshot has applied, so snapshot application can be handled by 1) flushing the apply thread, 2) sending the `MsgStorageAppend` that contains a snapshot to the `LocalAppendThread` to be applied.

## Usage

When asynchronous storage writes is enabled, the responsibility of code using the library is different from what is presented in [raft/doc.go](https://github.com/etcd-io/etcd/blob/main/raft/doc.go) (which has been updated to include a section about async storage writes). Users still read from the `Node.Ready()` channel. However, they process the updates it contains in a different manner. Users no longer consult the HardState, Entries, and Snapshot fields (steps 1 and 3 in doc.go). They also no longer call `Node.Advance()` to indicate that they have processed all entries in the Ready (step 4 in doc.go). Instead, all local storage operations are also communicated through messages present in the Ready.Message slice.

The local storage messages come in two flavors. The first flavor is log append messages, which target a LocalAppendThread and carry Entries, HardState, and a Snapshot. The second flavor is entry application messages, which target a LocalApplyThread and carry CommittedEntries. Messages to the same target must be reliably processed in order. Messages to different targets can be processed in any order. Each local storage message carries a slice of response messages that must delivered after the corresponding storage write has been completed.

With Asynchronous Storage Writes enabled, the total state machine handling loop will look something like this:

```go
for {
	select {
	case <-s.Ticker:
		n.Tick()
	case rd := <-s.Node.Ready():
		for _, m := range rd.Messages {
			switch m.To {
			case raft.LocalAppendThread:
				toAppend <- m
			case raft.LocalApplyThread:
				toApply <-m
			default:
				sendOverNetwork(m)
			}
		}
	case <-s.done:
		return
	}
}
```

Usage of Asynchronous Storage Writes will typically also contain a pair of storage handler threads, one for log writes (append) and one for entry application to the local state machine (apply). Those will look something like:

```go
// append thread
go func() {
	for {
		select {
		case m := <-toAppend:
			saveToStorage(m.State, m.Entries, m.Snapshot)
			send(m.Responses)
		case <-s.done:
			return
		}
	}
}

// apply thread
go func() {
	for {
		select {
		case m := <-toApply:
			for _, entry := range m.CommittedEntries {
				process(entry)
				if entry.Type == raftpb.EntryConfChange {
					var cc raftpb.ConfChange
					cc.Unmarshal(entry.Data)
					s.Node.ApplyConfChange(cc)
				}
			}
			send(m.Responses)
		case <-s.done:
			return
		}
	}
}
```

## Compatibility

The library remains backwards compatible with existing users and the change does not introduce any breaking changes. Users that do not set `AsyncStorageWrites` to true in the `Config` struct will not notice a difference with this change. This is despite the fact that the existing "synchronous storage writes" interface was adapted to share a majority of the same code. For instance, `Node.Advance` has been adapted to transparently acknowledge an asynchronous log append attempt and an asynchronous state machine application attempt, internally using the same message passing mechanism introduced in this change.

The change has no cross-version compatibility concerns. All changes are local to a process and nodes using asynchronous storage writes appear to behave no differently from the outside. Clusters are free to mix nodes running with and without asynchronous storage writes.

## Performance

The bulk of the performance evaluation of this functionality thus far has been done with [rafttoy](https://github.com/nvanbenschoten/rafttoy), a benchmarking harness developed to experiment with Raft proposal pipeline optimization. The harness can be used to run single-node benchmarks or multi-node benchmarks. It supports plugable raft logs, storage engines, network transports, and pipeline implementations.

To evaluate this change, we fixed the raft log (`etcd/wal`), storage engine (`pebble`), and network transport (`grpc`). We then built (https://github.com/nvanbenschoten/rafttoy/pull/3) a pipeline implementation on top of the new asynchronous storage writes functionality and compared it against two other pipeline implementations.

The three pipeline implementations we compared were:
- **basic** (**P1**): baseline stock raft usage, similar to the code in `doc.go`
- **parallel append + early ack** (**P2**): CockroachDB's current pipeline, which includes two significant variations to the basic pipeline. The first is that it sends MsgApp messages to followers before writing to local Raft log (see [commit](https://github.com/cockroachdb/cockroach/commit/b67eb696f6a064d6a3a18517c1ed5525faa931e8) for explanation), allowing log appends to occur in parallel across replicas. The second is that it acknowledges committed log entries before applying them (see [commit](https://github.com/cockroachdb/cockroach/commit/87aaea70d68903ad04ecae1adf2910333aa07a78) for explanation).
- **async append + async apply + early ack** (**P3**): A pipelining using asynchronous storage writes with a separate append thread and a separate apply thread. Also uses the same early acknowledgement optimization from above to ack committed entries before handing them to the apply thread.

All testing was performed on a 3 node AWS cluster of m5.4xlarge instances with gp3 EBS volumes (16000 IOPS, 1GB/s throughput). The testing used an open-loop workload to increase the rate of new raft proposals until a saturation point was reached.

Throughput vs latency of Raft proposal pipeline implementations

The comparison demonstrates two different benefits of asynchronous storage writes.

The first is that it reduces end-to-end latency of proposals by 20-25%. For instance, when serving 16MB/s of write traffic, **P1**'s average latency was 13.2ms, **P2**'s average latency was 7.3ms, and **P3**'s average latency was 5.2ms. This is a reduction in average latency of **29%** from the optimized pipeline that does not use asynchronous storage writes. This matches the expectations outlined in https://github.com/cockroachdb/cockroach/issues/17500.

The second is that it increases the maximum throughput at saturation. In this test, **P1** and **P2** topped out at 30MB/s, while **P3** could push up to 52MB/s, an increase in maximum throughput of **73%**. This is because asynchronous storage writes can improve batching for both log appends and log application. In this experiment, we saw the average append batch size under saturation increase from 928 to 1542, which is a similar ratio to the increase in peak throughput. We see a similar difference for apply batch sizes.

An additional performance comparison was presented in https://github.com/etcd-io/etcd/pull/14627#issuecomment-1302846128.

cc. @tbg @bdarnell


raft.go
```diff
@@ -793,7 +901,7 @@ func (r *raft) becomeLeader() {
 	// uncommitted log quota. This is because we want to preserve the
 	// behavior of allowing one entry larger than quota if the current
 	// usage is zero.
-	r.reduceUncommittedSize([]pb.Entry{emptyEnt})
+	r.uncommittedSize = 0
```
tbg, line 904: Circling back on https://github.com/etcd-io/etcd/pull/14627#discussion_r1021299292

Isn't this line unnecessary in the first place? The empty entry is zero-sized and we did call `r.reset()`. Let's add a unit test instead verifying that `payloadsSize(emptyEntry) == 0` and be done with it.

I filed https://github.com/etcd-io/raft/issues/11 for tracking uncommitted size across leadership changes.

raftpb/raft.proto
```diff
@@ -71,22 +75,32 @@ message Message {
...
+	// (type=MsgStorageAppendResp,index=100,logTerm=5) means the local node wrote
+	// entries up to index=100 in stable storage, and the term of the entry at index
+	// 100 was 5.
 	optional uint64      logTerm     = 5  [(gogoproto.nullable) = false];
 	optional uint64      index       = 6  [(gogoproto.nullable) = false];
 	repeated Entry       entries     = 7  [(gogoproto.nullable) = false];
 	optional uint64      commit      = 8  [(gogoproto.nullable) = false];
+	// hardState can be non-nil for MsgStorageAppend messages and is nil for all
+	// other message types.
+	optional HardState   hardState   = 13 [(gogoproto.nullable) = true];
```
tbg, line 91: Where did we land on https://github.com/etcd-io/etcd/pull/14627#discussion_r1015477471?

rawnode.go
```diff
@@ -131,7 +136,261 @@ func (rn *RawNode) Ready() Ready {
...
+// storage.
+func newStorageAppendRespMsg(r *raft, rd Ready) pb.Message {
+	m := pb.Message{
+		Type: pb.MsgStorageAppendResp,
+		To:   r.id,
+		From: LocalAppendThread,
+		// Dropped after term change, see below.
+		Term: r.Term,
+	}
+	if r.raftLog.hasNextOrInProgressUnstableEnts() {
```
tbg, line 271: TODO(tbg): review this bit more

raftpb/raft.proto
```diff
@@ -71,22 +75,32 @@ message Message {
 	optional uint64      from        = 3  [(gogoproto.nullable) = false];
 	optional uint64      term        = 4  [(gogoproto.nullable) = false];
 	// logTerm is generally used for appending Raft logs to followers. For example,
-	// (type=MsgApp,index=100,logTerm=5) means leader appends entries starting at
-	// index=101, and the term of entry at index 100 is 5.
+	// (type=MsgApp,index=100,logTerm=5) means the leader appends entries starting
+	// at index=101, and the term of the entry at index 100 is 5.
 	// (type=MsgAppResp,reject=true,index=100,logTerm=5) means follower rejects some
 	// entries from its leader as it already has an entry with term 5 at index 100.
+	// (type=MsgStorageAppendResp,index=100,logTerm=5) means the local node wrote
```
tbg, line 82: Is this entirely correct? In `newStorageAppendRespMsg`, we're putting the lastIndex/lastTerm only conditionally. Also, it's the lastIndex/lastTerm, not that of the entries passed in.

doc.go
```diff
@@ -169,6 +169,94 @@ given ID MUST be used only once even if the old node has been removed.
...
+fields (steps 1 and 3 above). They also no longer call Node.Advance() to
+indicate that they have processed all entries in the Ready (step 4 above).
+Instead, all local storage operations are also communicated through messages
+present in the Ready.Message slice.
+
+The local storage messages come in two flavors. The first flavor is log append
+messages, which target a LocalAppendThread and carry Entries, HardState, and a
+Snapshot. The second flavor is entry application messages, which target a
+LocalApplyThread and carry CommittedEntries. Messages to the same target must be
+reliably processed in order.Messages to different targets can be processed in
```
tbg, line 193: `order.Messages`

log.go
```diff
@@ -252,20 +327,51 @@ func (l *raftLog) commitTo(tocommit uint64) {
...
+func (l *raftLog) acceptApplying(i uint64, size entryEncodingSize, allowUnstable bool) {
+	if l.committed < i {
+		l.logger.Panicf("applying(%d) is out of range [prevApplying(%d), committed(%d)]", i, l.applying, l.committed)
+	}
+	l.applying = i
+	l.applyingEntsSize += size
+	// Determine whether to pause entry application until some progress is
+	// acknowledged. We pause in two cases:
+	// 1. the outstanding entry size equals or exceeds the maximum size.
+	// 2. the outstanding entry size does not equal the maximum size, but we
```
tbg, line 354: does not exceed

log.go
```diff
@@ -252,20 +327,51 @@ func (l *raftLog) commitTo(tocommit uint64) {
...
+	//    determine that the next entry in the log will push us over the limit.
+	//    We determine this by comparing the last entry returned from
+	//    raftLog.nextCommittedEnts to the maximum entry that the method was
+	//    allowed to return had there been no size limit. If these indexes are
+	//    not equal, then the returned entries slice must have been truncated to
+	//    adhere to the memory limit.
+	l.applyingEntsPaused = l.applyingEntsSize >= l.maxApplyingEntsSize ||
+		i < l.maxAppliableIndex(allowUnstable)
```
tbg, line 362: Does the second condition imply the first? In other words, can we remove the first part?

testdata/async_storage_writes_append_aba_race.txt
```diff
@@ -0,0 +1,493 @@
...
+
+raft-log 1
+----
+1/11 EntryNormal ""
+2/12 EntryNormal ""
+
+deliver-msgs 1
+----
+AppendThread->1 MsgStorageAppendResp Term:2 Log:2/12
+INFO 1 [term: 3] ignored entry appends from a MsgStorageAppendResp message with lower term [term: 2]
```
tbg, line 446: 👍🏽

rawnode.go
```diff
@@ -194,14 +194,22 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
 	return entsnum != 0 || st.Vote != prevst.Vote || st.Term != prevst.Term
 }
 
-func needStorageAppend(rd Ready, haveMsgsAfterAppend bool) bool {
+func needStorageAppendMsg(r *raft, rd Ready) bool {
```
tbg, line 197: This isn't the right place for this comment but I didn't want to post it in the main thread (no "Resolve" feature there):

I wonder if there's a simple UX improvement we could make to ready handling when using async writes. Right now, there are a bunch of fields that are unused in that case:

- MustSync
- CommittedEntries
- Entries
- Snapshot
- any others?

What about a helper method

```go
func AsyncStorageWritesReady(rd Ready) AsyncReady {
  // ...
}
```

that basically returns only the things you need to care about as a user of the async interface? This would also provide a convenient place to anchor documentation without having to interleave the two different usage patterns.

log.go
```diff
@@ -36,18 +36,30 @@ type raftLog struct {
...
+	// or asynchronously).
 	// Invariant: applied <= committed
 	applied uint64
 
 	logger Logger
 
-	// maxNextCommittedEntsSize is the maximum number aggregate byte size of the
-	// messages returned from calls to nextCommittedEnts.
-	maxNextCommittedEntsSize uint64
+	// maxApplyingEntsSize is the maximum outstanding byte size of the messages
```
tbg, line 52: The comment read slightly ambiguous to me. I thought it was tracking the maximum apply batch we had seen during the lifetime of the instance, and why we would need that had me confused.

I then realized this is simply the limit. Maybe clarify this for dummies like me

```suggestion
// maxApplyingEntsSize limits the outstanding byte size of the messages
```

tbg (approved):

tbg:
Getting close! No substantial comments.
I'd classify my comment about an AsyncReady struct is semi-substantial. I wouldn't want it to necessarily block the merge here, but I think there is a significant improvement to be made here in how accessible this new functionality it is to other adopters. So I consider it important that we discuss it and follow it to its conclusion in a timely manner.

log.go
nvanbenschoten: Not quite. It is possible for `i == l.maxAppliableIndex(allowUnstable)` but for the first condition to be true in two different ways:
- `l.applyingEntsSize == l.maxApplyingEntsSize`: The appliable entries in the log were exactly `maxApplyingEntsSize`
- `l.applyingEntsSize >  l.maxApplyingEntsSize`: The first appliable entry in the log was larger than `maxApplyingEntsSize`

I spent some time trying to avoid needing to add both conditions, but I don't see how to eliminate either of them.

doc.go
nvanbenschoten: Done.

raftpb/raft.proto
nvanbenschoten: I think it is what we want to say though. The delivery of a `MsgStorageAppendResp` with `index=100,logTerm=5` means that entries up to index=100/term=5 have been written in stable storage. It doesn't necessarily mean that they were written by the corresponding `MsgStorageAppend` though. I added a sentence to call that out.

rawnode.go
nvanbenschoten: I like this idea. Let me try it out and see how it looks.

log.go
nvanbenschoten: Done.

raft.go
nvanbenschoten: Thanks for opening the issue, and good point about this being unnecessary.

I switched this back to `r.reduceUncommittedSize(payloadSize(emptyEnt))` for now to make fewer assumptions. I also opened https://github.com/etcd-io/raft/issues/12 to remove the line entirely. Mind slapping `good first issue` and `help wanted` labels on it?

raftpb/raft.proto
nvanbenschoten: I had added a TODO to `newStorageAppendMsg` to revisit this suggestion. Let me do that now. I agree that we'll want to resolve it before merging because this is part of the external interface of the library.

nvanbenschoten: Done. I think I like it, but let me know what you think about the new commit at the end (`store Vote in MsgStorageAppend, not *HardState`).

tbg:
Can't review the diffs (no fixups?) but I looked at the one last new commit, still LGTM.

You may want to merge this and pursue https://github.com/etcd-io/raft/pull/8#discussion_r1054117426 separately.

tbg:
Merging with @nvanbenschoten's consent.

rawnode.go
```diff
@@ -131,7 +136,261 @@ func (rn *RawNode) Ready() Ready {
...
+		// If async storage writes are enabled, enqueue messages to
+		// local storage threads, where applicable.
+		if needStorageAppendMsg(r, rd) {
+			m := newStorageAppendMsg(r, rd)
+			rd.Messages = append(rd.Messages, m)
+		}
+		if needStorageApplyMsg(rd) {
+			m := newStorageApplyMsg(r, rd)
+			rd.Messages = append(rd.Messages, m)
+		}
```
ahrtr, line 170: When async write is enabled, you get `rd.Entries, rd.HardState and rd.Snapshot` encapsulated in `MsgStorageAppend` message. But you do not clear the fields, so you expect application ignore the fields, and only process the `MsgStorageAppend` message? 

The same for the `MsgStorageApply` message.

PR #17, opened 2022-12-30
raft: disabling uncommitted size accounting for emptyEntry

### LINKED ISSUE #12: remove unnecessary uncommitted size reduction after leadership changes
Related to https://github.com/etcd-io/raft/issues/11.

In `becomeLeader`, we have the following logic:
https://github.com/etcd-io/raft/blob/91eb7518afb589ac295f9fa3b837a11c48cc8682/raft.go#L792-L796

This is a no-op. The payload size of `emptyEnt` is 0, so there's no need to subtract it from the uncommitted size accounting.

If we delete the line, all tests pass. Let's delete it and then add a unit test verifying that `payloadsSize(emptyEntry) == 0` with a suitable comment explaining why this is an important property.

daman1807:
In becomeLeader method(raft.go) we reduce the uncommitted size for empty entry, which is basically a no-op as payload size will always be 0 for empty entries. Added a test case to check that as well.

Signed-off-by: Daman Arora <daman1807@github.com>

this closes #12

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -897,11 +897,10 @@ func (r *raft) becomeLeader() {
 		// This won't happen because we just called reset() above.
 		r.logger.Panic("empty entry was dropped")
 	}
-	// As a special case, don't count the initial empty entry towards the
-	// uncommitted log quota. This is because we want to preserve the
-	// behavior of allowing one entry larger than quota if the current
-	// usage is zero.
-	r.reduceUncommittedSize(payloadSize(emptyEnt))
+	// The payloadSize of an empty entry is 0 (see TestPayloadSizeOfEmptyEntry),
+	// so the preceding log append does not count against the uncommitted log
+	// quota of the new leader. In other words, after the call to appendEntry,
+	// r.uncommittedSize is still 0.
 	r.logger.Infof("%x became leader at term %d", r.id, r.Term)
 }
 
--- a/util_test.go
+++ b/util_test.go
@@ -139,3 +139,11 @@ func TestIsResponseMsg(t *testing.T) {
 		}
 	}
 }
+
+// TestPayloadSizeOfEmptyEntry ensures that payloadSize of empty entry is always zero.
+// This property is important because new leaders append an empty entry to their log,
+// and we don't want this to count towards the uncommitted log quota.
+func TestPayloadSizeOfEmptyEntry(t *testing.T) {
+	e := pb.Entry{Data: nil}
+	require.Equal(t, 0, int(payloadSize(e)))
+}


raft_test.go
```diff
@@ -5030,3 +5030,10 @@ func newTestRawNode(id uint64, election, heartbeat int, storage Storage) *RawNod
 	}
 	return rn
 }
+
+// TestPayloadSizeOfEmptyEntry ensures that payloadSize of empty entry is always zero
+func TestPayloadSizeOfEmptyEntry(t *testing.T) {
```
nvanbenschoten, line 5035: Let's move this to `util_test.go`, so that it lives next to the `payloadSize` utility function in `util.go`.

raft_test.go
```diff
@@ -5030,3 +5030,10 @@ func newTestRawNode(id uint64, election, heartbeat int, storage Storage) *RawNod
 	}
 	return rn
 }
+
+// TestPayloadSizeOfEmptyEntry ensures that payloadSize of empty entry is always zero
```
nvanbenschoten, line 5034: Comments should have punctuation, per the go style guide. So period at the end.

Also, could you add a second sentence that explains why this is an important property?
> This property is important because new leaders append an empty entry to their log and we don't want this to count towards the uncommitted log quota.

raft_test.go
```diff
@@ -5030,3 +5030,10 @@ func newTestRawNode(id uint64, election, heartbeat int, storage Storage) *RawNod
...
+func TestPayloadSizeOfEmptyEntry(t *testing.T) {
+	if payloadSize(pb.Entry{Data: nil}) != 0 {
+		t.Fatalf("EmptyEntry expected size: %d, got: %d", 0, payloadSize(pb.Entry{Data: nil}))
```
nvanbenschoten, line 5037: We've been moving the library over to use the `testify/require` library. So let's adjust this slightly:
```go
e := pb.Entry{Data: nil}
require.Equal(t, 0, payloadSize(e))
```

raft.go
```diff
@@ -901,7 +901,8 @@ func (r *raft) becomeLeader() {
 	// uncommitted log quota. This is because we want to preserve the
 	// behavior of allowing one entry larger than quota if the current
 	// usage is zero.
-	r.reduceUncommittedSize(payloadSize(emptyEnt))
```
nvanbenschoten, line 904: > // As a special case ...

Now that this is a no-op (it wasn't always), this isn't exactly a special case. I think we can remove the original comment and the commented-out line of code and say something simpler like:
> The payloadSize of an empty entry is 0 (see TestPayloadSizeOfEmptyEntry), so the preceding log append does not count against the uncommitted log quota of the new leader. In other words, after the call to appendEntry, r. uncommittedSize is still 0.

nvanbenschoten (commented):
Thanks for the PR @daman1807.

daman1807:
@nvanbenschoten I have pushed the required changes

daman1807:
@nvanbenschoten I can see the usage of golang standard testing library for assertions at lot of places, should I create a new issue/PR that uses testify/require for assertions?

nvanbenschoten (approved):
Thanks @daman1807. This looks good to me after the one remaining comment is addressed. Could you then squash the commits together?

We'll then want @tbg to sign off and merge.

> I can see the usage of golang standard testing library for assertions at lot of places, should I create a new issue/PR that uses testify/require for assertions?

@tbg what do you think? Worth the code movement, or should we just update tests as they're changed going forward? @falser101 had started doing some of this in https://github.com/etcd-io/raft/commit/cdd4cb38ef696c51ef44aefac8acafd302ac34e1 and related commits.

daman1807:
@nvanbenschoten I have updated the comment and squashed all the commits.

nvanbenschoten (approved):

daman1807:
Hi @tbg
Can I start picking test cases which need to be updated to use **testify/require** ?

> @tbg what do you think? Worth the code movement, or should we just update tests as they're changed going forward? @falser101 had started doing some of this in https://github.com/etcd-io/raft/commit/cdd4cb38ef696c51ef44aefac8acafd302ac34e1 and related commits.

tbg:
> Can I start picking test cases that need to be updated to use testify/require?

Yes, please! Please mention me (@tbg) on these PRs so they pop up on my radar.

I'll merge this once it gets through CI, thank you for your contribution!

tbg (approved):

PR #21, opened 2023-01-10
security: add dependabot.yml

ahrtr:
FYI. https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts

Signed-off-by: Benjamin Wang <wachao@vmware.com>


spzala (approved):
lgtm Thanks @ahrtr

PR #31, opened 2023-03-07
raft: fix panic on MsgApp after log truncation

pavelkalinnikov:
This PR fixes a panic and adds a regression test for the following scenario:

1. The flow of `MsgApp` from the leader to a follower is throttled (i.e. `Inflights` is full).
2. The leader doesn't fetch entries from storage and only periodically sends `MsgApp`s with empty `Entries`.
3. A log compaction/truncation happens in the background, and cuts out a portion of the log beyond what's still in-flight towards the slow follower (i.e. `Progress.Match` < log cutoff)
4. Some messages to the slow follower get dropped, and as a result it replies with a rejection `MsgAppResp`.
5. The leader resets `Progress.Next = Progress.Match+1`, and is about to retry sending entries from this point.
6. In `raft.maybeSendAppend` it calls `raftLog.term` and gets 0 for the missing entry (instead of some indication/error that the log was truncated at this index). It also skips fetching entries (as in step 2), and goes ahead sending an empty `MsgApp` (with `LogTerm = 0` and a fresh `Commit` index).
7. When the follower gets this `MsgApp`, in `raftLog.maybeAppend` it a) wrongly passes the `matchTerm` check because the 0 index matches the 0 corresponding to a missing entry in the local log, b) tries to bump the `Commit` index and panics because this index is beyond its local log's `lastIndex()`.

This bug was introduced in 42419da5. Specifically, the steps (2) and (6) previously used to unconditionally fetch `raftLog.entries()`, which would return `ErrCompacted` in the above scenario, and prevent sending the problematic `MsgApp`. The commit above introduced a condition under which the `ErrCompacted` would be unnoticed.

This PR makes `maybeSendAppend` more aware of this compaction scenario, and prevents sending the problematic `MsgApp`.

### DIFF
--- a/log.go
+++ b/log.go
@@ -153,7 +153,7 @@ func (l *raftLog) findConflict(ents []pb.Entry) uint64 {
 		if !l.matchTerm(ne.Index, ne.Term) {
 			if ne.Index <= l.lastIndex() {
 				l.logger.Infof("found conflict at index %d [existing term: %d, conflicting term: %d]",
-					ne.Index, l.zeroTermOnErrCompacted(l.term(ne.Index)), ne.Term)
+					ne.Index, l.zeroTermOnOutOfBounds(l.term(ne.Index)), ne.Term)
 			}
 			return ne.Index
 		}
@@ -388,11 +388,14 @@ func (l *raftLog) term(i uint64) (uint64, error) {
 		return t, nil
 	}
 
-	// The valid term range is [index of dummy entry, last index].
-	dummyIndex := l.firstIndex() - 1
-	if i < dummyIndex || i > l.lastIndex() {
-		// TODO: return an error instead?
-		return 0, nil
+	// The valid term range is [firstIndex-1, lastIndex]. Even though the entry at
+	// firstIndex-1 is compacted away, its term is available for matching purposes
+	// when doing log appends.
+	if i+1 < l.firstIndex() {
+		return 0, ErrCompacted
+	}
+	if i > l.lastIndex() {
+		return 0, ErrUnavailable
 	}
 
 	t, err := l.storage.Term(i)
@@ -444,7 +447,7 @@ func (l *raftLog) matchTerm(i, term uint64) bool {
 }
 
 func (l *raftLog) maybeCommit(maxIndex, term uint64) bool {
-	if maxIndex > l.committed && l.zeroTermOnErrCompacted(l.term(maxIndex)) == term {
+	if maxIndex > l.committed && l.zeroTermOnOutOfBounds(l.term(maxIndex)) == term {
 		l.commitTo(maxIndex)
 		return true
 	}
@@ -515,11 +518,11 @@ func (l *raftLog) mustCheckOutOfBounds(lo, hi uint64) error {
 	return nil
 }
 
-func (l *raftLog) zeroTermOnErrCompacted(t uint64, err error) uint64 {
+func (l *raftLog) zeroTermOnOutOfBounds(t uint64, err error) uint64 {
 	if err == nil {
 		return t
 	}
-	if err == ErrCompacted {
+	if err == ErrCompacted || err == ErrUnavailable {
 		return 0
 	}
 	l.logger.Panicf("unexpected error (%v)", err)
--- a/raft.go
+++ b/raft.go
@@ -571,7 +571,12 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 		return false
 	}
 
-	term, errt := r.raftLog.term(pr.Next - 1)
+	lastIndex, nextIndex := pr.Next-1, pr.Next
+	lastTerm, errt := r.raftLog.term(lastIndex)
+	if lastIndex != 0 && lastTerm == 0 && errt == nil {
+		errt = ErrCompacted
+	}
+
 	var ents []pb.Entry
 	var erre error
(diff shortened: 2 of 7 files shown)


pavelkalinnikov:
@nvanbenschoten @tbg PTAL

pavelkalinnikov:
@ahrtr PTAL

raft.go
```diff
@@ -571,7 +571,12 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 		return false
 	}
 
-	term, errt := r.raftLog.term(pr.Next - 1)
+	index := pr.Next - 1
```
nvanbenschoten, line 574: I like the change to stop using `pr.Next` multiple times, because it's not clear when it changes or what derived indexes from the value mean (e.g. why `pr.Next - 1`?).

Should we switch the call to `raftLog.entries` to use `index + 1`? Or better, should we split `index` into `lastIndex := pr.Next - 1` and `nextIndex := pr.Next` and replace both uses of `pr.Next` (`raftLog.entries`, `UpdateOnEntriesSend`) below? The use of both indexes looks like a bug today. I'd imagine named variables would help readers understand this better.

log.go
```diff
@@ -389,10 +389,11 @@ func (l *raftLog) term(i uint64) (uint64, error) {
 	}
 
 	// The valid term range is [index of dummy entry, last index].
-	dummyIndex := l.firstIndex() - 1
-	if i < dummyIndex || i > l.lastIndex() {
-		// TODO: return an error instead?
-		return 0, nil
+	if i+1 < l.firstIndex() {
```
nvanbenschoten, line 392: Out of curiosity, what would happen if we just removed this logic and let the code call `storage.Term`?

log.go
```diff
@@ -389,10 +389,11 @@ func (l *raftLog) term(i uint64) (uint64, error) {
 	}
 
 	// The valid term range is [index of dummy entry, last index].
```
nvanbenschoten, line 391: We no longer have a "dummy entry" variable, so it's not clear what this comment is referring to. FWIW, I never find it clear when reading this code why `i == l.firstIndex() - 1` is valid. Perhaps you could use this as a chance to clarify.

log.go
pavelkalinnikov: I tried this, tests fail. This is probably due to the fact that `l.firstIndex() != storage.FirstIndex()`. The former takes into account the in-memory `unstable` portion of the log which patches the storage state.

raft.go
pavelkalinnikov: Done.

log.go
pavelkalinnikov: Done.

nvanbenschoten (approved):
LGTM

nvanbenschoten (approved):

nvanbenschoten (approved):

nvanbenschoten:
Github isn't liking my approval. It may be because I'm not a maintainer.

rafttest/interaction_env_logger.go
```diff
@@ -82,17 +82,23 @@ func (l *RedirectLogger) Errorf(format string, v ...interface{}) {
 
 func (l *RedirectLogger) Fatal(v ...interface{}) {
 	l.print(4, v...)
+	panic(v)
```
ahrtr, line 85: Two comments:
1. Previously it just prints "`FATAL`" , and not to really panicking; so you changed the behavior;
2. `panic` accepts a parameter of type `interface{}`, but the type of v is `[]interface{}`.  You will get the panicking message something like below,
```
panic: ([]interface {}) 0xc00000c030

goroutine 1 [running]:
main.demo(...)
        /.../go/src/examples/varidicTest/main.go:7
main.main()
        /.../go/src/examples/varidicTest/main.go:11 +0x9d
exit status 2
```

It seems `panic(fmt.Sprint(v))` makes more sense than `panic(v)`? 

But I would suggest to  keep the existing behavior? cc @tbg

pavelkalinnikov: I basically mirrored the `nodeTestHarness` behaviour, see:
https://github.com/etcd-io/raft/blob/c728a6dece7abb918d8fe1c0b397646774028821/node_util_test.go#L61-L79

Point 2 is good, I fixed the panic argument in both loggers.

Re point 1:

It only changes the behaviour of the data-driven tests in `interaction_test.go`, which is the only user of this logger. The way it changes it is a real panic, so if there were tests relying on the old behaviour, it would break them. No tests were broken, so this behaviour wasn't relied upon and this change is ok.

The old behaviour was problematic, because in situations where the code wants to exit with a fatal or panic (e.g. in situations where critical invariants are broken), this logger would instead allow the code to proceed. As a result, there is a high risk that the code will break/fatal/panic in some other place. While preparing this PR, I ran into this exact situation: the code called `logger.Panicf`, proceeded, and ran into a real `panic` elsewhere. Since the log printed into the `logger` is not reported by the datadriven package in such case, I wasn't aware of the first ("root cause") panic, and was looking for a problem elsewhere until eventually I located the earlier panic.

I think we do need to fix this behaviour, but the exact way we do it may vary. The current version of this PR turns it into a real `panic`, which mirrors the other logger in this repo. This seems ok as a quick fix. The longer-term fix should make the output of the `logger` more visible. I reflected it in the TODO in the code.

ahrtr (approved):
LGTM

Thanks @pavelkalinnikov

PR #32, opened 2023-03-08
raft: paginate the unapplied config changes scan

pavelkalinnikov:


raft.go
```diff
@@ -927,14 +927,27 @@ func (r *raft) hasUnappliedConfChanges() bool {
...
-	if err != nil {
-		r.logger.Panicf("unexpected error getting unapplied entries (%v)", err)
-	}
-	for i := range ents {
-		if ents[i].Type == pb.EntryConfChange || ents[i].Type == pb.EntryConfChangeV2 {
-			return true
+	// Scan all unapplied committed entries to find a config change. Paginate the
+	// scan, to avoid a potentially unlimited memory spike.
+	//
+	// TODO(pavelkalinnikov): this scan does not need to happen each time. It's
```
pavelkalinnikov, line 933: OTOH, this scan happens rarely (only before election), so overall it might be cheaper (and simpler) to do a rare enough scan than maintain this predicate through every state change. Maybe I should replace this TODO with a note why the scan is ok.

pavelkalinnikov: There is a `raft.pendingConfIndex` field which might be already tracking what we need here.

raft.go
```diff
@@ -914,19 +914,44 @@ func (r *raft) hup(t CampaignType) {
...
+		}
+		for i := range ents {
+			if ents[i].Type == pb.EntryConfChange || ents[i].Type == pb.EntryConfChangeV2 {
+				return true
+			}
+		}
+		idx += uint64(len(ents))
+	}
+	return false
+}
```
ahrtr, line 953: The pagination doesn't make sense. Actually the memory has already been allocated in `(*raftLog) slice`, and it just returns a slice on top a underlying array. 

I would suggest to simplify the method something like below,
```
func (r *raft) hasUnappliedConfChanges() bool {
	if r.raftLog.applied >= r.raftLog.committed { // in fact applied == committed
		return false
	}

	ents, err := r.raftLog.slice(r.raftLog.applied+1, r.raftLog.committed+1, noLimit)
	if err != nil {
		r.logger.Panicf("unexpected error getting unapplied entries from (%v)", err)
	}
	for i := range ents {
		if ents[i].Type == pb.EntryConfChange || ents[i].Type == pb.EntryConfChangeV2 {
			return true
		}
	}

	return false
}

```

ahrtr: > `raftLog.slice` fetches a prefix of entries [from storage](https://github.com/etcd-io/raft/blob/2e0653d20750c7cbbf8473fd1582c63a867996c1/log.go#L470-L471), and a suffix of entries from memory (in the `unstable` type). The storage fetching part can read arbitrary amount of data into memory if the `committed..applied` range is large, and/or if there are large (in bytes) entries in this range. Pagination protects from this scenario.

Right, the key point is that `(*raftLog) slice` might reallocate memory at [log.go#L490](https://github.com/etcd-io/raft/blob/7699239b357a3acf26860eb6683fdcc788b07349/log.go#L490). I am thinking we need to skip [log.go#L487-L497](https://github.com/etcd-io/raft/blob/7699239b357a3acf26860eb6683fdcc788b07349/log.go#L487-L497) if [ents](https://github.com/etcd-io/raft/blob/7699239b357a3acf26860eb6683fdcc788b07349/log.go#L485) already hits the `maxSize`. If `len(ents) < cap(ents)`, then it means it already hits the `maxSize`.  

We might also need to change

https://github.com/etcd-io/raft/blob/7699239b357a3acf26860eb6683fdcc788b07349/storage.go#L133

to 
```
ents := ms.ents[lo-offset : hi-offset: hi-offset]
```

tbg (commented):
This change is what I expected, I'm a bit surprised to see the discussion on it. Perhaps the two of you can clue me in on what the discussion is about, I was mostly confused reading it, so am likely missing something.

raft.go
ahrtr: Can we merge https://github.com/etcd-io/raft/pull/34 firstly? Afterwards, rebase this PR, then we can take a look again.

pavelkalinnikov: @tbg I like the idea of `paginate` method. It also allows avoiding the `append` of the `unstable` in `raftLog.slice` - we can just return the `unstable` in its own page.

tbg: Here is the same experiment as above but with this PR. You can see that it solves the problem, memory usage doesn't spike at all.

image

ahrtr: Thanks for the explanation.  My original comment isn't valid anymore after merging a couple of other PRs.

ahrtr:
The PR looks good to me. Can you rebase this PR although there is no conflict? There are huge differences between this PR and the main branch.

raft.go
ahrtr: Agreed to discuss it separately.

pavelkalinnikov:
@ahrtr Sure, will rebase. Please don't merge yet, there is one suggested refactoring from @tbg to address.

raft.go
pavelkalinnikov: Added a `scan` method. Put the memory usage graphs to the PR description.

raft.go
```diff
@@ -922,33 +922,35 @@ func (r *raft) hup(t CampaignType) {
...
-		} else if len(ents) == 0 {
-			r.logger.Panicf("could not read unapplied entries in [%d, %d)", idx, end)
-		}
+	lo, hi := r.raftLog.applied+1, r.raftLog.committed+1
+	// Reuse the maxApplyingEntsSize limit because it is used for similar purposes
+	// (limiting the read of unapplied committed entries) when raft sends entries
+	// via the Ready struct for application.
+	// TODO(pavelkalinnikov): find a way to budget memory/bandwidth for this scan
+	// outside the raft package.
+	pageSize := r.raftLog.maxApplyingEntsSize
```
ahrtr, line 941: Question: what's the default value set in CockroachDB? I mean `MaxCommittedSizePerReady`. Does CockroachDB hard code this value or support users customizing it?

pavelkalinnikov: It is 64 MiB by default, configurable via an [env](https://github.com/cockroachdb/cockroach/blob/dd9e2a95fce4fcbc10ef8d36ff90b7286add9f1d/pkg/base/config.go#L290-L294) variable.

ahrtr: thx

log.go
```diff
@@ -461,6 +461,29 @@ func (l *raftLog) restore(s pb.Snapshot) {
...
 }
 
+// scan visits all log entries in the [lo, hi) range, returning them via the
+// given callback. The callback can be invoked multiple times, with consecutive
+// sub-ranges of the requested range. Returns up to pageSize bytes worth of
+// entries at a time. May return more if a single entry size exceeds the limit.
+//
+// If the callback returns an error, scan terminates and returns this error
+// immediately. This can be used to stop the scan early ("break" the loop).
+func (l *raftLog) scan(lo, hi uint64, pageSize entryEncodingSize, v func([]pb.Entry) error) error {
```
tbg, line 471: could add that the log entries `[lo,hi)` all have to exist. As is, one could be tempted to call `scan(1, math.MaxUint64, ...)` to scan the entire log, but this would return an error eventually.

PR #34, opened 2023-03-10
raft: improve log slice allocations

pavelkalinnikov:
This PR improves allocations in `raftLog.slice()` method:

- appends to the slice returned by `Storage.Entries` directly instead of copying, because the semantics allow to do so (the returned slice is owned by the caller). This sometimes may avoid an allocation depending on the storage implementation.
- allocates exactly as much as needed. Previously, the entire `unstable` log could be copied and then truncated.

### DIFF
--- a/log.go
+++ b/log.go
@@ -491,10 +491,7 @@ func (l *raftLog) slice(lo, hi uint64, maxSize entryEncodingSize) ([]pb.Entry, e
 	if hi > l.unstable.offset {
 		unstable := l.unstable.slice(max(lo, l.unstable.offset), hi)
 		if len(ents) > 0 {
-			combined := make([]pb.Entry, len(ents)+len(unstable))
-			n := copy(combined, ents)
-			copy(combined[n:], unstable)
-			ents = combined
+			ents = append(ents, unstable...)
 		} else {
 			ents = unstable
 		}
--- a/log.go
+++ b/log.go
@@ -510,7 +510,7 @@ func (l *raftLog) slice(lo, hi uint64, maxSize entryEncodingSize) ([]pb.Entry, e
 	}
 	// Otherwise, total size of unstable does not exceed maxSize-size, so total
 	// size of ents+unstable does not exceed maxSize. Simply concatenate them.
-	return append(ents, unstable...), nil
+	return extend(ents, unstable), nil
 }
 
 // l.firstIndex <= lo <= hi <= l.firstIndex + len(l.entries)
--- a/log.go
+++ b/log.go
@@ -463,40 +463,54 @@ func (l *raftLog) restore(s pb.Snapshot) {
 
 // slice returns a slice of log entries from lo through hi-1, inclusive.
 func (l *raftLog) slice(lo, hi uint64, maxSize entryEncodingSize) ([]pb.Entry, error) {
-	err := l.mustCheckOutOfBounds(lo, hi)
-	if err != nil {
+	if err := l.mustCheckOutOfBounds(lo, hi); err != nil {
 		return nil, err
 	}
 	if lo == hi {
 		return nil, nil
 	}
-	var ents []pb.Entry
-	if lo < l.unstable.offset {
-		storedEnts, err := l.storage.Entries(lo, min(hi, l.unstable.offset), uint64(maxSize))
-		if err == ErrCompacted {
-			return nil, err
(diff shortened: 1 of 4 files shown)


pavelkalinnikov:
@tbg @ahrtr PTAL. This is built on top of #45, so the first 4 commits can be skipped in this review. I will rebase when the other PR is merged.

log.go
```diff
@@ -463,42 +463,57 @@ func (l *raftLog) restore(s pb.Snapshot) {
...
+	if uint64(len(ents)) < cut-lo {
+		return ents, nil
+	}
+	if len(ents) == 1 && entryEncodingSize(ents[0].Size()) >= maxSize {
+		return ents, nil
+	}
+
+	// Slow path computes the actual total size, so that unstable entries are cut
+	// optimally before being copied to ents slice.
+	size := entsSize(ents)
```
tbg, line 504: So here we know that we have `cut-lo` entries were read from storage, right? Doesn't this imply that `size <= maxSize`? So is the next condition only true if the existing slice exactly hits `maxSize`?

tbg: Do we have a test that hits the conditional?

tbg (approved):

ahrtr:
@pavelkalinnikov could you rebase this PR since the base PR https://github.com/etcd-io/raft/pull/45 has already been merged?

pavelkalinnikov:
@tbg @ahrtr This is ready for review.

util.go
```diff
@@ -318,3 +318,20 @@ func assertConfStatesEquivalent(l Logger, cs1, cs2 pb.ConfState) {
...
 	l.Panic(err)
 }
+
+// extend appends vals to the given dst slice. It differs from the standard
+// slice append only in the way it allocates memory. If cap(dst) is not enough
+// for appending the values, precisely size len(dst)+len(vals) is allocated.
+//
+// Use this instead of standard append in situations when this is the last
+// append to dst, so there is no sense in allocating more than needed.
+func extend[T any](dst, vals []T) []T {
```
tbg, line 328: This will make raft/v3 incompatible with go 1.17 and under, should we avoid that by monomorphizing explicitly?

cc @ahrtr probably a general discussion to be had about which min Go version to target.

tbg (changes requested):
Only question is about `extend`.

util.go
pavelkalinnikov: Done.

log.go
```diff
@@ -463,43 +463,57 @@ func (l *raftLog) restore(s pb.Snapshot) {
...
+	if size >= maxSize {
+		return ents, nil
+	}
```
ahrtr, line 507:
```suggestion
	// Slow path computes the actual total size, so that unstable entries are cut
	// optimally before being copied to ents slice.
	size := entsSize(ents)
	if size >= maxSize {
		return ents, nil
	}
```

util.go
ahrtr: Good to see the change is reverted.

log.go
pavelkalinnikov: Nice one.

I also replaced `entryEncodingSize(unstable[0].Size())` below with just `entsSize(unstable)`, for readability.

log.go
pavelkalinnikov: After the change @ahrtr suggested, this condition can now be hit both ways: `size` can be > `maxSize` if `len(ents) == 1` and the first entry is large.

log.go
```diff
@@ -463,43 +463,54 @@ func (l *raftLog) restore(s pb.Snapshot) {
...
+
+	unstable := limitSize(l.unstable.slice(l.unstable.offset, hi), maxSize-size)
+	// Total size of unstable may exceed maxSize-size only if len(unstable) == 1.
+	// If this happens, ignore this extra entry.
+	if len(unstable) == 1 && size+entsSize(unstable) > maxSize {
+		return ents, nil
 	}
```
ahrtr, line 510: Can save some cpu cycles? 
```suggestion
	if len(unstable) == 1 {
	    if size+entsSize(unstable) > maxSize {
		return ents, nil
	    }
	}
```

ahrtr: pls ignore this comment, there is no difference from your implementation.

ahrtr (approved):
LGTM

Thanks @pavelkalinnikov

tbg (approved):

PR #36, opened 2023-03-14
raft: don't panic when looking for term conflicts

pavelkalinnikov:
This commit fixes a hypothetical panic that may occur when a stale `MsgApp`
message arrives to a follower. The conflict searching algorithm in
`findConflictByTerm` may return a log index which is not present in the log, and
thus the `raftLog.term()` method may return an error. It is safe to ignore this
error and send `MsgAppResp` with the found index and a zero `LogTerm`.

This commit also restores the behaviour that existed before commit d0fb0cd6.
Back then, the `term()` function would silently return 0 instead of an error, and
a zero `LogTerm` would be sent with the rejecting `MsgAppResp`. After that commit,
there is a new possible panic. We remove the possibility of this panic here.

### DIFF
--- a/log.go
+++ b/log.go
@@ -162,34 +162,31 @@ func (l *raftLog) findConflict(ents []pb.Entry) uint64 {
 	return 0
 }
 
-// findConflictByTerm takes an (index, term) pair (indicating a conflicting log
-// entry on a leader/follower during an append) and finds the largest index in
-// log l with a term <= `term` and an index <= `index`. If no such index exists
-// in the log, the log's first index is returned.
+// findConflictByTerm returns a best guess on where this log ends matching
+// another log, given that the only information known about the other log is the
+// (index, term) of its single entry.
 //
-// The index provided MUST be equal to or less than l.lastIndex(). Invalid
-// inputs log a warning and the input index is returned.
-func (l *raftLog) findConflictByTerm(index uint64, term uint64) uint64 {
-	if li := l.lastIndex(); index > li {
-		// NB: such calls should not exist, but since there is a straightfoward
-		// way to recover, do it.
-		//
-		// It is tempting to also check something about the first index, but
-		// there is odd behavior with peers that have no log, in which case
-		// lastIndex will return zero and firstIndex will return one, which
-		// leads to calls with an index of zero into this method.
-		l.logger.Warningf("index(%d) is out of range [0, lastIndex(%d)] in findConflictByTerm",
-			index, li)
-		return index
-	}
-	for {
-		logTerm, err := l.term(index)
-		if logTerm <= term || err != nil {
-			break
+// Specifically, the first returned value is the max guessIndex <= index, such
+// that term(guessIndex) <= term or term(guessIndex) is not known (because this
+// index is compacted or not yet stored).
+//
+// The second returned value is the term(guessIndex), or 0 if it is unknown.
+//
+// This function is used by a follower and leader to resolve log conflicts after
+// an unsuccessful append to a follower, and ultimately restore the steady flow
+// of appends.
+func (l *raftLog) findConflictByTerm(index uint64, term uint64) (uint64, uint64) {
+	for ; index > 0; index-- {
+		// If there is an error (likely ErrCompacted or ErrUnavailable), we don't
+		// know whether it's a match or not, so assume a possible match and return
+		// the index, with 0 term indicating an unknown term.
+		if ourTerm, err := l.term(index); err != nil {
+			return index, 0
+		} else if ourTerm <= term {
+			return index, ourTerm
 		}
-		index--
 	}
-	return index
+	return 0, 0
 }
 
 // nextUnstableEnts returns all entries that are available to be written to the
--- a/log.go
+++ b/log.go
@@ -107,24 +107,25 @@ func (l *raftLog) String() string {
 // maybeAppend returns (0, false) if the entries cannot be appended. Otherwise,
 // it returns (last index of new entries, true).
 func (l *raftLog) maybeAppend(index, logTerm, committed uint64, ents ...pb.Entry) (lastnewi uint64, ok bool) {
-	if l.matchTerm(index, logTerm) {
-		lastnewi = index + uint64(len(ents))
-		ci := l.findConflict(ents)
-		switch {
-		case ci == 0:
-		case ci <= l.committed:
-			l.logger.Panicf("entry %d conflict with committed entry [committed(%d)]", ci, l.committed)
-		default:
-			offset := index + 1
-			if ci-offset > uint64(len(ents)) {
-				l.logger.Panicf("index, %d, is out of range [%d]", ci-offset, len(ents))
-			}
-			l.append(ents[ci-offset:]...)
-		}
-		l.commitTo(min(committed, lastnewi))
-		return lastnewi, true
+	if !l.matchTerm(index, logTerm) {
+		return 0, false
 	}
-	return 0, false
+
+	lastnewi = index + uint64(len(ents))
+	ci := l.findConflict(ents)
+	switch {
+	case ci == 0:
+	case ci <= l.committed:
+		l.logger.Panicf("entry %d conflict with committed entry [committed(%d)]", ci, l.committed)
+	default:
+		offset := index + 1
+		if ci-offset > uint64(len(ents)) {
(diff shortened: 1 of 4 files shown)


log.go
```diff
@@ -182,12 +189,13 @@ func (l *raftLog) findConflictByTerm(index uint64, term uint64) uint64 {
 			index, li)
 		return index
 	}
-	for {
+	for ; ; index-- {
 		logTerm, err := l.term(index)
 		if logTerm <= term || err != nil {
+			// NB: an error can only be ErrCompacted here, but it's always safe to
+			// return the current index, so we don't assert on other errors.
```
tbg, line 196: Yes and no because the method comment claims that this returns the "best possible" index but in the case of a transient storage error this isn't true. Could we return an error here if we see one that is unexpected (i.e. not ErrCompacted).

raft.go
```diff
@@ -1667,9 +1667,13 @@ func (r *raft) handleAppendEntries(m pb.Message) {
 		hintIndex := min(m.Index, r.raftLog.lastIndex())
 		hintIndex = r.raftLog.findConflictByTerm(hintIndex, m.LogTerm)
 		hintTerm, err := r.raftLog.term(hintIndex)
-		if err != nil {
-			panic(fmt.Sprintf("term(%d) must be valid, but got %v", hintIndex, err))
+		if err != nil { // can be ErrCompacted or ErrUnavailable
```
tbg, line 1670: something like `panicUnlessErrCompactedOrErrUnavailable(err)`?

raft.go
```diff
@@ -1667,9 +1667,13 @@ func (r *raft) handleAppendEntries(m pb.Message) {
 		hintIndex := min(m.Index, r.raftLog.lastIndex())
 		hintIndex = r.raftLog.findConflictByTerm(hintIndex, m.LogTerm)
```
tbg, line 1668: Wouldn't it be better for this to return the term, since we're just about to recompute it?

raft.go
```diff
@@ -1667,9 +1667,13 @@ func (r *raft) handleAppendEntries(m pb.Message) {
...
 		hintTerm, err := r.raftLog.term(hintIndex)
-		if err != nil {
-			panic(fmt.Sprintf("term(%d) must be valid, but got %v", hintIndex, err))
+		if err != nil { // can be ErrCompacted or ErrUnavailable
+			hintTerm = 0
 		}
+		// NB: findConflictByTerm always returns an index <= its argument, so
+		// hintIndex <= min(m.Index, lastIndex()). This guarantees that the next
+		// append will be at a lower index. This is also the reason why it is safe
+		// to ignore a possible error above which, though, should never happen.
```
tbg, line 1676: I'm still uneasy about not populating the term hint. The reason the leader can handle this is because originally it was never populated, but we should consider this legacy behavior we'd want to migrate out of if we had a way to do so. If `findConflictByTerm` returned `(index, term, error)` and on success both `(index,term)` are populated, there would be fewer questions. And on error, we could log a warning and simply not respond, as is admissible in raft, and in particular for a rejection, or we could respond with the unadjusted `(index,term)`, so something like this:

```go
hintIndex := ...
hintTerm, err := ....
if err != nil {
  return // don't respond, log warning
}
adjHintIndex, adjHintTerm, err := r.raftLog.findConflictByTerm(...)
if err != nil { log warning but continue } else {
hintIndex, hintTerm = adjHintIndex, adjHintTerm
}
// respond
```

tbg (approved):
Looks good to me, thanks! I left some suggestions to a have fewer "unexpected" return values and no case in which the term is not populated. The goal would be to remove all edge cases and make it more transparent what happens in the truncation scenarios, etc.

It might also be good to add some targeted unit testing for `findConflictByTerm`. I saw that you added a test case in `TestFastLogRejection`, but a few more wouldn't hurt. I should have added that right when we introduced `findConflictByTerm`, sorry about that.

Approving preemptively because the current code is "good enough" and I'm fairly certain it's correct (mod comments). The improvements would make it easier to be convinced that it's correct in a month or two but I'll rely on your judgment of which further improvements are worth it.

log.go
```diff
@@ -164,15 +164,22 @@ func (l *raftLog) findConflict(ents []pb.Entry) uint64 {
...
+// log l with index <= `index` and term <= `term`.
 //
 // The index provided MUST be equal to or less than l.lastIndex(). Invalid
 // inputs log a warning and the input index is returned.
+//
+// If index < firstIndex()-1 then the information about the term at this index
+// has been compacted away. In this case the input index is returned. Similarly,
+// if all terms for entries in [firstIndex()-1, index] are greater than the
+// given term, i.e. the possible suitable entry has been compacted, then
+// firstIndex()-2 is returned. Note that these two cases probably indicate a
```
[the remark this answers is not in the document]
pavelkalinnikov: We would return 0 still, because `raftLog.term(0) == 0`. But I refactored this func to guard against `index` underflow, and also made the method comment better.

log.go
```diff
@@ -182,12 +189,13 @@ func (l *raftLog) findConflictByTerm(index uint64, term uint64) uint64 {
 			index, li)
 		return index
 	}
-	for {
+	for ; ; index-- {
```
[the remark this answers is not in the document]
pavelkalinnikov: With the next iteration of this PR, this loop is now `for ; index > 0; index-- {` which is slightly more justified. Leaving as is.

log.go
pavelkalinnikov: This comment is softer now. Decided not to return an err / overengineer here, because `term()` method currently panics on any error other than `ErrCompacted/ErrUnavailable`. This all feels clunky and not ideal, maybe error handling in this library should be improved.

But also: returning both a higher and a lower hint here is fine - if it doesn't work, there will be another MsgApp iteration, or there will be more entries resent than necessary. The `MsgApp` sender (leader) works in such a way that it can swallow any hint including zero.

raft.go
pavelkalinnikov: Done.

raft.go
pavelkalinnikov: Moved away from this error check.

raft.go
pavelkalinnikov: As discussed in chat, it's more important to send a rejection `MsgAppResp` than get the hint exactly right.

There is an implicit expectation that a flow of `MsgApp` messages will eventually lead to a `MsgResp` (potentially with a reject). This is sort of a liveness bump for appends. If we introduce a condition under which a rejection `MsgResp` is not sent, we add some risk of ending up in a busy loop on the leader.

raft.go
```diff
@@ -1909,6 +1909,8 @@ func (r *raft) abortLeaderTransfer() {
 
 // committedEntryInCurrentTerm return true if the peer has committed an entry in its term.
 func (r *raft) committedEntryInCurrentTerm() bool {
+	// NB: r.Term is never 0 on a leader, so if zeroTermOnOutOfBounds returns 0,
```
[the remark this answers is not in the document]
pavelkalinnikov: Yes, I manually checked that this func is only called on a leader. But it's also somewhat intuitive: the leader checks that it committed entries during its current term.

raft_test.go
```diff
@@ -4703,72 +4705,52 @@ func TestFastLogRejection(t *testing.T) {
...
-				t.Errorf("peer 1 step append fail")
-			}
+			require.NoError(t, n1.Step(msgs[0]), "peer 1 step append fail")
 			msgs = n1.readMessages()
-			if msgs[0].LogTerm != test.nextAppendTerm {
-				t.Fatalf("#%d expected log term = %d, but got %d", i, test.nextAppendTerm, msgs[0].LogTerm)
-			}
-			if msgs[0].Index != test.nextAppendIndex {
-				t.Fatalf("#%d expected index = %d, but got %d", i, test.nextAppendIndex, msgs[0].Index)
-			}
```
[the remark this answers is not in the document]
pavelkalinnikov: Removed.

pavelkalinnikov:
@tbg Addressed comments, and added unit tests for `findConflictByTerm`.

pavelkalinnikov:
CC @ahrtr: This removes a panic condition. Please merge since this is now approved, or feel free to review too.

@tbg I modified this PR after approval, but it's equivalent. Feel free to take a final look though, and please merge if you see this earlier.

Thank you.

Issue #43, opened 2023-04-11
stepWait not used in ProposeConfChange


lavacat:
`stepWait` was [implemented](https://github.com/etcd-io/etcd/pull/9137) to fail-fast in [Propose](https://github.com/etcd-io/raft/blob/main/node.go#L444) method.
But it was never used in [ProposeConfChange](https://github.com/etcd-io/raft/blob/main/node.go#L469) method.

[Question](https://github.com/etcd-io/etcd/pull/9137#issuecomment-474218191) was raised about it on original PR.
Also, it's was causing timeout in ExampleCluster_memberAddAsLearner test. [comment](https://github.com/etcd-io/etcd/issues/14040#issuecomment-1323254093)

This isn't a critical issue, because in production ProposeConfChange isn't usually called consecutively. But it seams to me that `stepWait` was missed for `ProposeConfChange` method.

folays:
OP wonders about `ProposeConfchange()`.

Besides `ProposeConfChange()`,  `Propose()`  has also some problems which probably stems from the same origins.

I will talk only about cluster where `DisableProposalForwarding = false`, because `= true` has it own set of unfollow-ability of failed proposals. It helps drafting things quickly, but it's bad in ensuring reliability.

`(*node).Propose()` method is widly undocumented on what can it be expected of it.
An API is like a contract, and without specifications on it, what are its limitation etc..., it's like there is no guarantee at all to have it to eventually do something in any specific timeframe, so it's unuable in any serious production environment without first reading all code path it branches into to figure out those.

`.Propose()` only specification is _`Note that proposals can be lost without notice, therefore it is user's job to ensure proposal retries.`_ which does not help at all. No words on when/how the loss can be noticed and/or assumed.

It appears that the set of predicates could hold for `.Propose()` :
- when it has been called on a current leader (`rd.SoftState.Lead == r.id`)
- and `.Propose()` returned `err == nil`
  - that means that it is guaranteed to be proposed and ultimately commited/applied...
- ...guarantee which would hold as long as "this" leader consider itself to have leadership
  - otherwise you should trigger cancellation/timeout of the ongoing proposal waiting status
- which would mean we seem to be able to assume that the Proposal does not need retrial in the meantime if the rest of the conditions hold

`.ProposeConfChange()` is worse, and its API behavior for callers seems to be "hope for the best", "retry after some unspecified time if commit does not appears in `.CommittedEntries[]`"

[...]

PR #44, opened 2023-04-11
raft: error on ignored proposal and wait for EntryConfChange results

lavacat:
fixes: https://github.com/etcd-io/raft/issues/43


lavacat:
there might be some other side effects of returning `ErrProposalDropped`. If there is interest in merging this PR, I can do some more digging/testing.

raft.go
```diff
@@ -1229,7 +1229,7 @@ func stepLeader(r *raft, m pb.Message) error {
 
 				if refused != "" {
 					r.logger.Infof("%x ignoring conf change %v at config %s: %s", r.id, cc, r.prs.Config, refused)
-					m.Entries[i] = pb.Entry{Type: pb.EntryNormal}
```
fuweid, line 1232: Switching to normal entry is trying to increase applied index so that the next confChange can be applicable. It seems that it is right move to return dropped

chaochn47: > Switching to normal entry is trying to increase applied index so that the next confChange can be applicable.

I don't think switching to normal entry will make a difference here. But I agree returning proposal dropped early is helpful.

raft.go
```diff
@@ -1229,7 +1229,7 @@ func stepLeader(r *raft, m pb.Message) error {
...
+					return ErrProposalDropped
```
chaochn47, line 1232: I don't know the side effect of dropping other entries in the same `m.Entries` for loop. 

How about moving it til the end of stepLeader function, especially after `bcastAppend` so other entries handling are not impacted just to be safe?

chaochn47:
etcd only uses so-called deprecated `pb.ConfChange`, but I am wondering how this change will impact `pb.ConfChangeV2` handling. 

https://github.com/etcd-io/raft/blob/3cbcd6e7088c773cab6331459162561682e0d1df/node.go#L141-L153

cc @ahrtr @serathius for insights~

raft.go
fuweid: You are right. The raft will update applied index even if the apply fails. I was thinking that the normal is used to catch up with pendingConfIndex so that it won't impact the following confChange.

chaochn47: aha. Thanks for explanation! I think next `ProposeConfChange()` will succeed once the `raft.Applied` is advanced by etcd raftNode `r.Advance()`

ahrtr:
> etcd only uses so-called deprecated `pb.ConfChange`, but I am wondering how this change will impact `pb.ConfChangeV2` handling.

raft will automatically convert it to V2, see [node.go#L533](https://github.com/etcd-io/raft/blob/ffef590a6132c1e4da2572dacd32afb7abf988dd/node.go#L533).

> [m.Entries[i] = pb.Entry{Type: pb.EntryNormal}](https://github.com/etcd-io/raft/blob/ffef590a6132c1e4da2572dacd32afb7abf988dd/raft.go#L1231-L1232)

The intention is just to ignore the confChange, as the log "`ignoring conf change`" mentions.

raft.go
lavacat: the loop is only processing config changes, see `if cc != nil {`. So, I think it's ok to drop here. Also, my understanding is that we have to drop full proposal, there is no per entry `ErrProposalDropped`.

chaochn47: I see, thanks for the explanation!

chaochn47:
> etcd only uses so-called deprecated `pb.ConfChange`, but I am wondering how this change will impact pb.ConfChangeV2 handling. 

To be clear, I am not sure what's the impact of fail fast in case of `pb.ConfChangeV2` handling https://github.com/etcd-io/raft/blob/7357439323abcd0654b7146022f7a58793d667eb/raft.go#L1222-L1226

Is there an unknown reason that designs like this? @tbg for consulting..

raft.go
lavacat: For reference, the use of normal entry was added [here](https://github.com/etcd-io/raft/commit/cb095791f970bd0a2017785680de3fd440524059) 
I think the idea was to support mixed list of entries (ConfChanges and Normal) in proposal, but is it even possible to get mixed list right now? 
When I check for creation of `pb.Message{Type: pb.MsgProp`, there are two distinct places:
1. regular [Propose](https://github.com/etcd-io/raft/blob/65a0bf3b6779e6b84d6c032bde1fcdf7a49d13df/node.go#L444)
2. [confChangeToMsg](https://github.com/etcd-io/raft/blob/65a0bf3b6779e6b84d6c032bde1fcdf7a49d13df/node.go#L456)  - used in [ProposeConfChange](https://github.com/etcd-io/raft/blob/65a0bf3b6779e6b84d6c032bde1fcdf7a49d13df/node.go#L464) and [raft.appliedTo](https://github.com/etcd-io/raft/blob/5fe1c31c515840dd807b8f07d91f704f8334a1b6/raft.go#L696)

lavacat:
cc @nvanbenschoten, noticed some of your recent comments here

https://github.com/etcd-io/raft/blob/7357439323abcd0654b7146022f7a58793d667eb/raft.go#L698-L704

can you take a look at this PR?

nvanbenschoten:
@lavacat thanks for the ping. I don't see any issue with the changes made to `stepLeader`. There's now a decent amount of coupling between the rejection logic in `stepLeader` and the auto-leave proposal logic in `appliedTo`, but it appears to be correct.

PR #45, opened 2023-04-11
raft: clarify [Memory]Storage entries semantics

pavelkalinnikov:
This change adds extra protection to slices returned from `MemoryStorage`, so that any appends to these slices at the raft or application level can be tolerated.

This change also clarifies the semantics of `Storage.Entries` method to this extent, and touches other aspects like returned errors.

### DIFF
--- a/storage.go
+++ b/storage.go
@@ -130,8 +130,11 @@ func (ms *MemoryStorage) Entries(lo, hi, maxSize uint64) ([]pb.Entry, error) {
 		return nil, ErrUnavailable
 	}
 
-	ents := ms.ents[lo-offset : hi-offset]
-	return limitSize(ents, entryEncodingSize(maxSize)), nil
+	ents := limitSize(ms.ents[lo-offset:hi-offset], entryEncodingSize(maxSize))
+	// NB: use the full slice expression to limit what the caller can do with the
+	// returned slice. For example, an append will reallocate and copy this slice
+	// instead of corrupting the neighbouring ms.ents.
+	return ents[:len(ents):len(ents)], nil
 }
 
 // Term implements the Storage interface.
--- a/storage.go
+++ b/storage.go
@@ -279,8 +279,9 @@ func (ms *MemoryStorage) Append(entries []pb.Entry) error {
 	offset := entries[0].Index - ms.ents[0].Index
 	switch {
 	case uint64(len(ms.ents)) > offset:
-		ms.ents = append([]pb.Entry{}, ms.ents[:offset]...)
-		ms.ents = append(ms.ents, entries...)
+		// NB: full slice expression protects ms.ents at index >= offset from
+		// rewrites, as they may still be referenced from outside MemoryStorage.
+		ms.ents = append(ms.ents[:offset:offset], entries...)
 	case uint64(len(ms.ents)) == offset:
 		ms.ents = append(ms.ents, entries...)
 	default:
--- a/storage.go
+++ b/storage.go
@@ -242,7 +242,10 @@ func (ms *MemoryStorage) Compact(compactIndex uint64) error {
 	}
 
 	i := compactIndex - offset
-	ents := make([]pb.Entry, 1, 1+uint64(len(ms.ents))-i)
+	// NB: allocate a new slice instead of reusing the old ms.ents. Entries in
+	// ms.ents are immutable, and can be referenced from outside MemoryStorage
+	// through slices returned by ms.Entries().
+	ents := make([]pb.Entry, 1, uint64(len(ms.ents))-i)
 	ents[0].Index = ms.ents[i].Index
 	ents[0].Term = ms.ents[i].Term
 	ents = append(ents, ms.ents[i+1:]...)
--- a/storage.go
+++ b/storage.go
@@ -48,10 +48,28 @@ type Storage interface {
 
 	// InitialState returns the saved HardState and ConfState information.
 	InitialState() (pb.HardState, pb.ConfState, error)
-	// Entries returns a slice of log entries in the range [lo,hi).
-	// MaxSize limits the total size of the log entries returned, but
-	// Entries returns at least one entry if any.
+
+	// Entries returns a slice of consecutive log entries in the range [lo, hi),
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
@tbg Could you take a look commit-by-commit?

storage.go
```diff
@@ -48,10 +48,30 @@ type Storage interface {
...
+	//
+	// The raft library or application can append to the returned slice though.
+	// The Storage implementation must protect its state from corruption that such
+	// appends may cause. Two common ways to do so:
+	//  - allocate the slice before returning it, and assume that slice ownership
+	//  is transferred to this library and the application (the safest option),
+	//  - use Go full slice expressions before returning the slice, so that
+	//  appends will reallocate/copy it (see MemoryStorage).
+	//
+	// Returns ErrCompacted if entry lo has been compacted, or ErrUnavailable if
```
pavelkalinnikov, line 71: How accurate is this? I'm seeing that in `MemoryStorage`, `ErrUnavailable` is only returned if we haven't got any entries after the snapshot. But we also verify that `hi <= ms.lastIndex()+1`, and so is `lo`, so technically it seems that we return `ErrUnavailable` only if the requested `lo == hi`. This shouldn't be the only case though.

We should clarify these semantics, and fix `MemoryStorage` to be a reference implementation.

pavelkalinnikov:
@ahrtr PTAL

storage.go
```diff
@@ -48,10 +48,30 @@ type Storage interface {
...
 	InitialState() (pb.HardState, pb.ConfState, error)
-	// Entries returns a slice of log entries in the range [lo,hi).
-	// MaxSize limits the total size of the log entries returned, but
-	// Entries returns at least one entry if any.
+
+	// Entries returns a slice of log entries in the range [lo, hi). The maxSize
+	// limits the total size of the log entries returned, but Entries returns at
+	// least one entry if any.
+	//
+	// Entries that storage returns via this method are not modified by the raft
```
tbg, line 56: It would help to be more opinionated here. As a user, I don't want various options on how it could work (at least not when it's complicated like that) but just be told what to do.
We likely want to say that the caller of `Entries` owns the slice and must not mutate it (and that `raft` adheres to that rule). We should then say something similar in the `Ready` struct, i.e. the application must not modify the slices handed to it by raft. According to the above it would be up to `raft` to not append into the slice when there's spare capacity.

storage.go
tbg: https://github.com/cockroachdb/cockroach/blob/8f975693228f7a676204594540973de9494c54fb/pkg/kv/kvserver/logstore/logstore.go#L576-L585

https://github.com/cockroachdb/cockroach/blob/8f975693228f7a676204594540973de9494c54fb/pkg/kv/kvserver/logstore/logstore.go#L600-L601

So in CRDB, we get `ErrUnavailable` only if we're trying to scan `[lo,hi)` and we're finding that not all of the entries we are trying to return are there. For example, if the log is 5,6,7 we would return this error from `[5,8)`, `[1,4)`, `[1,6)`.

The `MemoryStorage` will instead panic if `hi >= lastIndex` and so returns `ErrUnavailable` only if we start to scan from below the first log index.

For a reference impl it would probably be better to match the CRDB behavior, i.e. return `ErrUnavailable` either way - this _should_ be a dead code path (because why would raft read past the last index?) but it's easier if the implementer can just handle these cases the same way.

storage.go
```diff
@@ -242,7 +242,10 @@ func (ms *MemoryStorage) Compact(compactIndex uint64) error {
...
 
 	i := compactIndex - offset
-	ents := make([]pb.Entry, 1, 1+uint64(len(ms.ents))-i)
+	// NB: allocate a new slice instead of reusing the old ms.ents. Entries in
+	// ms.ents are immutable, and can be referenced from outside MemoryStorage
+	// through slices returned by ms.Entries().
+	ents := make([]pb.Entry, 1, uint64(len(ms.ents))-i)
 	ents[0].Index = ms.ents[i].Index
 	ents[0].Term = ms.ents[i].Term
 	ents = append(ents, ms.ents[i+1:]...)
```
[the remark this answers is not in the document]
tbg: The meta question is, why optimize this testing-only implementation?
One way to do so would be to allow a more streamlined `raft.Storage` implementation, where instead of a user re-implementing a bunch of logic every time, they would just plug their storage engine (with a much narrower API) into a struct that we provide (essentially `MemoryStorage` but wrapping this user-supplied object rather than a slice).

tbg (changes requested):

storage.go
pavelkalinnikov: Reworded, PTAL.

`Ready` already states that all fields are read-only, so we're good there w.r.t. the entries slices too.

storage.go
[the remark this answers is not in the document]
ahrtr: > The meta question is, why optimize this testing-only implementation?


It isn't test-only. It's used by etcd. Please see [server.go#L2114](https://github.com/etcd-io/etcd/blob/29e79717217f49c46654309b5fb74546f7062967/server/etcdserver/server.go#L2114)

> allow a more streamlined raft.Storage implementation, where instead of a user re-implementing a bunch of logic every time, they would just plug their storage engine (with a much narrower API) into a struct that we provide

It's a big topic, which definitely should be discussed separately in a dedicated session.

ahrtr (approved):
LGTM

Nice micro-optimization.

storage.go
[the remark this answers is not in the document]
tbg: > Couldn't get the idea. What narrower API are you thinking of?

Not worth discussing further here, but each user of raft needs to implement `raft.Storage` and thus needs to re-implement a bunch of subtle details regarding how `Entries()` has to work, etc.

Instead, most users probably have something that can write an entry and point read an entry, so unless they want something more bespoke that thing that they have should be enough to get a `raft.Storage` implementation in which all of the subtlety is provided by raft.

pavelkalinnikov:
@ahrtr @tbg Thank you both for the reviews! Could you merge please?

PR #46, opened 2023-04-13
raft: use a type-safe slice of log entries

pav-kv:
All `[]pb.Entry` slices that `raft` works with are contiguous valid log ranges, for which the following invariants hold:

- `entries[i+1].Index == entries[i].Index + 1`
- `entries[i+1].Term >= entries[i].Term`
- TODO: we could also verify that `Term > 0` and `Index > 0`; the only zero case is the "sentinel" `(0, 0)` entry.

Much of the code implicitly assumes this, and will behave arbitrarily if this isn't true. This PR adds a `LogRange` type which is a type-safe slice promising these invariants explicitly.

---

Additionally, entry slices are often used in conjunction with an `(index, term)` tuple which describes the entry that immediately precedes the slice, i.e.:

- `entries[0].Index == index + 1`
- `entries[0].Term >= term`

For example, the log append handling stack (starting from a `MsgApp` message handler all the way down to `unstable` and `Storage`) operates with such a construct, effectively a "conditional put".

We have seen incorrect (not adhering to this invariant) `MsgApp` being sent (#47), or received (https://github.com/etcd-io/etcd/issues/14735). A type-safe / verified entries "append" slice should reduce the chance of such bugs, or at least make them better discoverable.

This PR introduces such a type, `logAppend`, and refactors the `raftLog` type to take advantage of it. This also makes the code cleaner.

This type could evolve into "the only" way of storing and passing entry slices around. Effectively, the log (or its sub-range) is always an `(index, term)` indicating a compaction or other "watermark", plus a `[]pb.Entry` suffix of currently available entries. This seems true for `unstable` and `raftLog` types. Another example is the `MemoryStorage`.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -757,7 +757,10 @@ func (r *raft) reset(term uint64) {
 }
 
 func (r *raft) appendEntry(es ...pb.Entry) (accepted bool) {
-	li := r.raftLog.lastIndex()
+	li, lt := r.raftLog.tip()
+	if r.Term < lt {
+		r.logger.Panicf("%x appending out-of-order term: %d < %d", r.id, r.Term, lt)
+	}
 	for i := range es {
 		es[i].Term = r.Term
 		es[i].Index = li + 1 + uint64(i)
@@ -1724,6 +1727,16 @@ func (r *raft) restore(s pb.Snapshot) bool {
 		// state when this method is called.
 		r.logger.Warningf("%x attempted to restore snapshot as leader; should never happen", r.id)
 		r.becomeFollower(r.Term+1, None)
+		return false
+	}
+
+	// Another defense-in-depth: the follower is seeing a snapshot at a bigger
+	// term, but hasn't updated its own term.
+	if s.Metadata.Term > r.Term {
+		r.logger.Warningf("%x attempted to restore snapshot at term %d while being at earlier term %d; "+
+			"should transition to follower at a larger term first",
+			r.id, s.Metadata.Term, r.Term)
+		r.becomeFollower(s.Metadata.Term, None)
 		return false
 	}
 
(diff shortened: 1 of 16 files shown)


pavelkalinnikov:
@tbg @ahrtr PTAL, what do you think of the concept? This is inspired by https://github.com/etcd-io/etcd/issues/15595 and https://github.com/etcd-io/etcd/issues/15704 where entry slices are [suspect to be] invalid.

In the meantime I'm looking at the panic in `TestLearnerReceiveSnapshot`.

range.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"testing"
+
+	pb "go.etcd.io/raft/v3/raftpb"
+)
+
+// LogRange represents a range of contiguous log entries. This entries slice has
+// been verified to have the following properties:
+//   - entries[i+1].Index == entries[i].Index + 1
+//   - entries[i+1].Term >= entries[i].Term
```
ahrtr, line 27: Nice!

raft.go
```diff
@@ -1644,11 +1646,15 @@ func stepFollower(r *raft, m pb.Message) error {
 }
 
 func (r *raft) handleAppendEntries(m pb.Message) {
+	entries, err := VerifyLogRange(m.Index, m.LogTerm, m.Entries)
```
ahrtr, line 1649: It may cause reduce of performance. So suggest to only enable the verification in test by default; and disable it in production. 

We can introduce an environment variable to enable / disable it.  FYI. https://github.com/etcd-io/etcd/blob/main/client/pkg/verify/verify.go

ahrtr:
It's a good idea to me, but we should only enable the verification in test. See https://github.com/etcd-io/raft/pull/46#discussion_r1166137357

pavelkalinnikov:
Fixing the test failures in #47 - it's a pre-existing bug in tests, and a lack of invariant checks that lead to an incorrect `MsgApp` being sent.

PR #47, opened 2023-04-14
raft: fix out-of-order terms

pavelkalinnikov:
This commit adds a couple invariant checks which were silently ignored and resulted in incorrectly setup tests passing, and incorrect messages exchanged.

Property 1: for a MsgApp message `m`, `m.Entries[i].Term >= m.LogTerm`. Or, more generally, `m.Entries` is a slice of contiguous entries with a non-decreasing `Term`, and `m.Index+1 == m.Entries[0].Index && m.LogTerm <= m.Entries[0].Term`.

This property was broken in a few tests. The root cause was that leader appends out-of-order entries to its own log when it `becomeLeader()`. This was a result of incorrectly set up tests: they restored to a snapshot at `(index=11,term=11)`, but became leader at an earlier term 1. Then it was sending the following, obviously incorrect, `MsgApp`: `{Index=11,LogTerm=11,Entries={{Index=12,Term=1}}`.

The fix in tests is either going through a follower state at the correct term, by calling `becomeFollower(term, ...)`, or initializing from a correct `HardState` in storage.

Property 2: For a `MsgSnap` message `m`, `m.Term >= m.Snapshot.Metadata.Term`, because the leader doesn't know of any higher term than its own, and hence can't send a message with such an inversion.

This was broken in `TestRestoreFromSnapMsg`, and is now fixed.


pavelkalinnikov:
@tbg @ahrtr PTAL

raft.go
```diff
@@ -757,7 +757,10 @@ func (r *raft) reset(term uint64) {
 }
 
 func (r *raft) appendEntry(es ...pb.Entry) (accepted bool) {
-	li := r.raftLog.lastIndex()
+	li, lt := r.raftLog.tip()
+	if r.Term < lt {
```
lavacat, line 761: after the first successful `appendEntry` `lt == r.Term`, because `es[i].Term = r.Term` on line 765. So, I don't think this check should be here. It probably should happen at the end of `raft.newRaft()`.

raft.go
```diff
@@ -757,7 +757,10 @@ func (r *raft) reset(term uint64) {
...
+		r.logger.Panicf("%x appending out-of-order term: %d < %d", r.id, r.Term, lt)
```
lavacat, line 762: nit: you aren't checking Terms in `es`, so msg is a bit confusing. Maybe `appending out-of-order term` -> `last raft log entry has term greater then raft struct`

raft_test.go
```diff
@@ -1366,11 +1366,11 @@ func TestHandleHeartbeat(t *testing.T) {
 // TestHandleHeartbeatResp ensures that we re-send log entries when we get a heartbeat response.
 func TestHandleHeartbeatResp(t *testing.T) {
 	storage := newTestMemoryStorage(withPeers(1, 2))
+	storage.SetHardState(pb.HardState{Term: 3, Vote: 1, Commit: 3})
 	storage.Append([]pb.Entry{{Index: 1, Term: 1}, {Index: 2, Term: 2}, {Index: 3, Term: 3}})
```
lavacat, line 1370: are these entries necessary for the test? Maybe both `storage.SetHardState` and `storage.Append` can be removed to keep things simpler

raft.go
lavacat: how about using `warning` instead of `panic`?

lavacat (changes requested):

raft.go
ahrtr: +1

Issue #48, opened 2023-04-25
Cluster struggles to elect leader when all peers are started at the same time


ahrtr:
Please refer to https://github.com/etcd-io/etcd/issues/15784

CaojiamingAlan:
@ahrtr Please review #50, thanks

pavelkalinnikov:
Closing since #50 has been merged.

stevenengler:
Thanks @CaojiamingAlan for the patch!

PR #50, opened 2023-05-10
change math/rand to crypto/rand to break tie in elections

CaojiamingAlan:
This is to solve [https://github.com/etcd-io/raft/issues/48](url)
Change math/rand to crypto/rand so that even in emulations the random numbers are different for each server.
Result verified by the given sample:
```
[ec2-user@ip-172-31-38-170 ~]$ cat shadow.data/hosts/server3/etcd.1000.stdout | grep leader
2000-01-01T00:00:02.65Z	info	raft	v3@v3.0.0-20221201111702-eaa6808e1f7a/node.go:344	raft.node: d88930ebb1eeba8b elected leader 4ec99d67dc5f5788 at term 2
[ec2-user@ip-172-31-38-170 ~]$ cat shadow.data/hosts/server1/etcd.1000.stdout | grep leader
2000-01-01T00:00:02.65Z	info	raft	v3@v3.0.0-20221201111702-eaa6808e1f7a/node.go:344	raft.node: 4f7b593e8bbade70 elected leader 4ec99d67dc5f5788 at term 2
[ec2-user@ip-172-31-38-170 ~]$ cat shadow.data/hosts/server2/etcd.1000.stdout | grep leader
2000-01-01T00:00:02.6Z	info	raft	v3@v3.0.0-20221201111702-eaa6808e1f7a/raft.go:903	4ec99d67dc5f5788 became leader at term 2
2000-01-01T00:00:02.6Z	info	raft	v3@v3.0.0-20221201111702-eaa6808e1f7a/node.go:344	raft.node: 4ec99d67dc5f5788 elected leader 4ec99d67dc5f5788 at term 2
```

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -16,14 +16,14 @@ package raft
 
 import (
 	"bytes"
+	"crypto/rand"
 	"errors"
 	"fmt"
 	"math"
-	"math/rand"
+	"math/big"
 	"sort"
 	"strings"
 	"sync"
-	"time"
 
 	"go.etcd.io/raft/v3/confchange"
 	"go.etcd.io/raft/v3/quorum"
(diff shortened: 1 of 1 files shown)


ahrtr (approved):
LGTM

cc @tbg @pavelkalinnikov

tbg (approved):

tbg:
(I verified that `globalRand` is used only for the election timeout, i.e. it's not on any hot path).

PR #52, opened 2023-05-17
Move from StatePause->StateReplicate on heartbeat response when possible

tbg:
See individual commits. Essentially, when a fully caught-up follower was
reported unreachable, it'd transition to `StateProbe` but then wouldn't
recover from that via heartbeats (once they resumed).

This caused some issues in CRDB because we rely on the reported status
to reason about the safety of leadership changes, etc.

This PR makes it such that StateProbe resolves on its own: when the
leader hears back from the follower via a heartbeat, it sends an
empty MsgApp, and as response to this moves the follower back into
StateProbe.


testdata/heartbeat_resp_recovers_from_paused.txt
```diff
@@ -67,12 +67,24 @@ stabilize
...
+> 2 receiving messages
+  1->2 MsgApp Term:1 Log:1/11 Commit:11
+> 2 handling Ready
+  Ready MustSync=false:
+  Messages:
+  2->1 MsgAppResp Term:1 Log:0/11
+> 1 receiving messages
+  2->1 MsgAppResp Term:1 Log:0/11
 
 # Sad, replica is still tracked as probing.
```
pavelkalinnikov, line 83: Fixed, update the comment in the last commit?

testdata/heartbeat_resp_recovers_from_paused.txt
```diff
@@ -67,12 +67,24 @@ stabilize
...
 status 1
 ----
 1: StateReplicate match=11 next=12
-2: StateProbe match=11 next=12
+2: StateReplicate match=11 next=12
 3: StateReplicate match=11 next=12
 
 # Need to actually distribute a command to sort things out.
```
pavelkalinnikov, line 90: Fix this comment (in the last commit) to say that now the flow stays in the Replicate state?

raft.go
```diff
@@ -1460,7 +1464,7 @@ func stepLeader(r *raft, m pb.Message) error {
 		// empty append, allowing it to recover from situations in which all the
 		// messages that filled up Inflights in the first place were dropped. Note
 		// also that the outgoing heartbeat already communicated the commit index.
-		if pr.Match < r.raftLog.lastIndex() {
+		if pr.Match < r.raftLog.lastIndex() || pr.State != tracker.StateReplicate {
```
pavelkalinnikov, line 1467: Is it fine to allow the StateSnapshot here? Maybe test that too?

tbg (commented):
Updated, PTAL. There is a new commit touching up `TestLeaderAppResp` (in preparation for changing the outcome of one test case in the main commit).

The main commit has been changed to only change behavior for StateProbe (not StateSnapshot) with commentary in both cases on why StateSnapshot is not affected.

pavelkalinnikov (approved):

tbg:
TFTRs!

Issue #29, opened 2023-02-28
Allow recovering from some assertions


tbg:
Consider an assertion such as

https://github.com/etcd-io/raft/blob/d9907d6ac6baaebc3c9fd4e67acaa4154d2b3cd3/log.go#L324

of which there are various across the codebase.

Hitting this panic usually means that a follower has not upheld its durability guarantees.

Violating invariants like these is not great, but crashing might not be the user's first choice here. An app might prefer to keep going despite some amount of risk that a write was lost (which often it won't have been).

The way I would structure this is by introducing event-based logging:

Instead of a line like this

https://github.com/etcd-io/raft/blob/d9907d6ac6baaebc3c9fd4e67acaa4154d2b3cd3/log.go#L324

We'd have something like this (total strawman just to get the idea across)

```go
l.logger.Event(&CommitOutOfRangeIndex{Commit: tocommit, LastIndex: l.lastIndex()})
// Code here to actually handle the problem gracefully
...
```

where the default logger would panic but users could define a logger that would just log the event and keep going. We wouldn't have to make all events that are now panics recoverable at first but could allow this only for certain events like the one discussed here.

Extracted from https://github.com/etcd-io/raft/issues/25#issuecomment-1449055381_

CaojiamingAlan:
I am trying to solve the commit index regression problem first. I am now writing an interaction test to reproduce the problem first.

Issue #18, opened 2023-01-04
Better handling of detected invariant violations


tbg:
Reincarnation of https://github.com/etcd-io/etcd/issues/10166#issue-368252159.

While it's good that raft tries to detect invariant violations (for example follower regressing no an acked log position), more flexibility in how to react these events would be desirable.

This isn't something we should do lightly, but the current failure modes are unnecessarily catastrophic. The point of raft is to provide high availability, so a faulty follower, even if it technically broke the rules, shouldn't crash the leader.

halegreen:
Is it a  valid solution :  having a option to disable this panic and recover a node in-place when it reaches this state. `panic: tocommit(265625) is out of range [lastIndex(265624)]. Was the raft log corrupted, truncated, or lost`

halegreen:
hi, I'd like to take this challenge,  raft is new for me, so the solution may cost sometime for me..

zouyonghao:
Hi, as discussed in https://github.com/etcd-io/etcd/issues/13493 , a reproducible record for this issue is also available with a docker:

```
docker run -it --rm --cap-add=SYS_PTRACE --security-opt seccomp=unconfined zouyonghao/etcd-10166 bash
# rr replay /root/10166/rr_rec_2_0/ -a
```

pav-kv:
Currently, a node that notices the property violation, crashes itself. In many cases, though, the safety properties violation is not necessarily this node's fault. If any Byzantine effect happens, any node could be the culprit, and we shouldn't be trying to figure out which.

An example scenario:
1. 3/5 followers are misconfigured, and writes on them are not durable.
2. The leader commits an entry based on votes/accepts from these followers.
3. The cluster restarts.
4. Now, the cluster will evolve in different directions depending on who becomes the leader.
5. In particular, if one of the faulty 3/5 nodes is elected, it will try to override commits on the former leader node. The former leader will notice this as a violation. But it would make no much sense to crash this node, it behaves correctly and the offenders were the followers.

[...]

PR #53, opened 2023-05-17
Add AllowInvariantViolations config, commit index regression test and…

CaojiamingAlan:
Attempt to resolve part of https://github.com/etcd-io/etcd/issues/10166, https://github.com/etcd-io/raft/issues/18, and https://github.com/etcd-io/raft/issues/29.
IMHO it is hard to keep backwards compatible while introducing event-based log, since it will require the users to implement the new Event method in the log interface if they have implemented their own loggers. 
Therefore, this commit only tries to introduce a new config option AllowInvariantViolation. If enabled, raft will try to recover from the commit index regression problem described in the above issues, rather than panic directly. The approach is exactly the one used in https://github.com/etcd-io/raft/pull/25. This may not be suitable for more complex situations e.g. when leader transfer is happening, but is enough for simple case like the one in the test. 
I'm very willing to discuss more about this.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -257,6 +257,12 @@ type Config struct {
 	// logical clock from assigning the timestamp and then forwarding the data
 	// to the leader.
 	DisableProposalForwarding bool
+
+	// AllowInvariantViolations is true if we want to proceed even if
+	// some invariants are violated. If enabled, the logger will not panic,
+	// but just log the event and keep going. An example would be a commit index
+	// regression in a follower.
+	AllowInvariantViolations bool
 }
 
 func (c *Config) validate() error {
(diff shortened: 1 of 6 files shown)


raft.go
```diff
@@ -1684,6 +1703,24 @@ func (r *raft) handleAppendEntries(m pb.Message) {
...
 
 func (r *raft) handleHeartbeat(m pb.Message) {
+	if r.allowInvariantViolations {
+		// If the following commit panics, it means the commit id in the message is rejected by the
+		// log since it does not have those entries. Hitting this panic usually means that a follower
+		// has not upheld its durability guarantees. If allowInvariantViolations is enabled, we take
+		// a naive approach to recover from the panic: send a pb.MsgHeartbeatResp with a reject hint,
+		// forcing the leader to retransmit the entries.
+		defer func() {
+			if v := recover(); v != nil {
```
lavacat, line 1713: imo adding `recover` adds more complexity. https://github.com/etcd-io/raft/pull/25#issuecomment-1449055381 suggested different approach: `default logger would panic but users could define a logger that would just log the event and keep going`

PR #54, opened 2023-05-18
tracker: fix byte counter leak in Inflights tracker

pavelkalinnikov:
This change fixes a bug in the Inflights tracker. The reset() method did not zero the bytes counter, which could result in a quota "leak" and delayed or stalled MsgApp sends.

The reset() method is used when the replication flow changes state between Probe/Replicate/Snapshot. If reset() is not called at an appropriate moment, when Inflights.Full(), the bytes counter would stay over the budget and stall the flow.

The test added in this PR failed before, and passes after the change.

### DIFF
--- a/tracker/inflights.go
+++ b/tracker/inflights.go
@@ -137,6 +137,7 @@ func (in *Inflights) Count() int { return in.count }
 
 // reset frees all inflights.
 func (in *Inflights) reset() {
-	in.count = 0
 	in.start = 0
+	in.count = 0
+	in.bytes = 0
 }
--- a/tracker/inflights_test.go
+++ b/tracker/inflights_test.go
@@ -227,6 +227,28 @@ func TestInflightsFull(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


tbg (commented):
Good find!

tbg (approved):

pavelkalinnikov:
@ahrtr FYI, this bug shouldn't affect etcd (I assume you don't yet use the [MaxInflightBytes](https://github.com/etcd-io/raft/blob/177ef28aae851bb644181bf531793ee33a895c56/raft.go#L223) option).

ahrtr:
> @ahrtr FYI, this bug shouldn't affect etcd (I assume you don't yet use the [MaxInflightBytes](https://github.com/etcd-io/raft/blob/177ef28aae851bb644181bf531793ee33a895c56/raft.go#L223) option).

No, etcd doesn't use it at all.

ahrtr (approved):
Good catch.
thx

PR #68, opened 2023-06-01
add `RawNode.CampaignNow()`

erikgrinaker:
@pavelkalinnikov @tbg Submitting this for early feedback. We may end up not needing this. Will add tests and such later, if we decide we want this.

The corresponding CockroachDB change is https://github.com/cockroachdb/cockroach/pull/104189.

---

This patch adds `RawNode.CampaignNow()`, which can be used to bypass the CheckQuorum condition that votes are only granted if the voter hasn't heard from the leader in the past election timeout interval. PreVote is otherwise applied and respected as usual, i.e. prevotes are not granted if the candidate is behind on its log.

In CockroachDB, where we quiesce inactive ranges and don't tick them, this is useful when we have reason to believe the leader died while the range was quiesced (via separate failure detectors).


erikgrinaker:
@pavelkalinnikov @tbg I think we'll probably want this, so I've finished this up.

testdata/checkquorum.txt
```diff
@@ -0,0 +1,322 @@
...
+WARN 1 stepped down to follower since quorum is not active
+INFO 1 became follower at term 1
+
+stabilize
+----
+> 1 handling Ready
+  Ready MustSync=false:
+  Lead:0 State:StateFollower
+  Messages:
+  1->2 MsgHeartbeat Term:1 Log:0/0 Commit:11
```
tbg, line 166: Add a comment about why there are four rounds of heartbeats queued up here?

testdata/checkquorum.txt
```diff
@@ -0,0 +1,322 @@
...
+  3->1 MsgHeartbeatResp Term:1 Log:0/0
+  INFO 1 [term: 2] ignored a MsgHeartbeatResp message with lower term from 3 [term: 1]
+  3->1 MsgHeartbeatResp Term:1 Log:0/0
+  INFO 1 [term: 2] ignored a MsgHeartbeatResp message with lower term from 3 [term: 1]
+> 1 handling Ready
+  Ready MustSync=true:
+  HardState Term:2 Commit:11
+
+# Other nodes can now successfully campaign. Note that we haven't ticked 3, so
+# it won't grant votes.
```
tbg, line 227: You can highlight this by stabilizing 3 separately in the next step. That way, the salient bit won't be buried in a few pages of text.

testdata/prevote.txt
```diff
@@ -0,0 +1,387 @@
+# Tests that PreVote prevents a node that is behind on the log to obtain
+# prevotes and call an election. Also test that a node that is up-to-date on its
+# log can hold an election (the recent leader condition requires CheckQuorum).
+
+add-nodes 3 voters=(1,2,3) index=10 prevote=true
```
tbg, line 5: ditto about suppressing uninteresting setup via log-level.

testdata/prevote_checkquorum.txt
```diff
@@ -0,0 +1,314 @@
...
+INFO newRaft 3 [peers: [1,2,3], term: 0, commit: 10, applied: 10, lastindex: 10, lastterm: 1]
+
+campaign 1
+----
+INFO 1 is starting a new election at term 0
+INFO 1 became pre-candidate at term 0
+INFO 1 [logterm: 1, index: 10] sent MsgPreVote request to 2 at term 0
+INFO 1 [logterm: 1, index: 10] sent MsgPreVote request to 3 at term 0
+
+stabilize
```
tbg, line 25: ditto log-level

raft.go
```diff
@@ -1080,10 +1103,14 @@ func (r *raft) Step(m pb.Message) error {
 
 	switch m.Type {
 	case pb.MsgHup:
+		force := bytes.Equal(m.Context, forceCampaign)
+		if force && r.readOnly.option == ReadOnlyLeaseBased {
+			r.logger.Panic("can't force an election with ReadOnlyLeaseBased")
```
tbg, line 1108: Why panic? We can "just" ignore the MsgHup, can't we?

testdata/campaign_now_lease.txt
```diff
@@ -0,0 +1,118 @@
...
+INFO newRaft 3 [peers: [1,2,3], term: 0, commit: 2, applied: 2, lastindex: 2, lastterm: 1]
+
+campaign 1
+----
+INFO 1 is starting a new election at term 0
+INFO 1 became candidate at term 1
+INFO 1 [logterm: 1, index: 2] sent MsgVote request to 2 at term 1
+INFO 1 [logterm: 1, index: 2] sent MsgVote request to 3 at term 1
+
+stabilize
```
tbg, line 20: ditto about noise suppression

testdata/checkquorum.txt
```diff
@@ -320,3 +321,109 @@ stabilize
...
+
+# CampaignNow will allow an election even with an active leader.
+campaign-now 1
+----
+INFO 1 is starting a new election at term 3 (force)
+INFO 1 became candidate at term 4
+INFO 1 [logterm: 3, index: 12] sent MsgVote (force) request to 2 at term 4
+INFO 1 [logterm: 3, index: 12] sent MsgVote (force) request to 3 at term 4
+
+stabilize
```
tbg, line 333: \# 1 is an eligible leader, so it wins the election.

raft.go
erikgrinaker: We can, but this is a programming error. The caller should have checked this already and returned an error (see `CampaignNow()`). I'm fine with downgrading this to an error and dropping the message though.

erikgrinaker:
Downgrading to draft while we explore alternatives.

erikgrinaker:
>Once a quorum of nodes individually consider the leader dead, a candidate will be able to obtain prevotes despite a recent leader.

It actually turns out we don't need to do anything at all here. The recent leader condition only applies to followers, so if we explicitly call `Campaign()` and transition the node to pre-candidate then it will grant votes to other (pre-)candidates. This means that if a quorum of nodes choose to campaign they can overthrow a current leader despite `PreVote+CheckQuorum`, which is exactly what we want.

PR #70, opened 2023-06-02
add data-driven tests for PreVote and CheckQuorum

erikgrinaker:
cc @pavelkalinnikov @tbg 

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -4978,6 +4978,12 @@ func setRandomizedElectionTimeout(r *raft, v int) {
 	r.randomizedElectionTimeout = v
 }
 
+// SetRandomizedElectionTimeout is like setRandomizedElectionTimeout, but
+// exported for use by tests that are not in the raft package, using RawNode.
+func SetRandomizedElectionTimeout(r *RawNode, v int) {
+	setRandomizedElectionTimeout(r.raft, v)
+}
+
 func newTestConfig(id uint64, election, heartbeat int, storage Storage) *Config {
 	return &Config{
 		ID:              id,
--- a/rafttest/interaction_env_handler.go
+++ b/rafttest/interaction_env_handler.go
@@ -97,6 +97,14 @@ func (env *InteractionEnv) Handle(t *testing.T, d datadriven.TestData) string {
 		// following, etc.). The information for node n is based on
 		// n's view.
 		err = env.handleRaftState()
+	case "set-randomized-election-timeout":
+		// Set the randomized election timeout for the given node. Will be reset
+		// again when the node changes state.
+		//
+		// Example:
+		//
+		// set-randomized-election-timeout 1 timeout=5
+		err = env.handleSetRandomizedElectionTimeout(t, d)
 	case "stabilize":
 		// Deliver messages to and run process-ready on the set of IDs until
 		// no more work is to be done. If no ids are given, all nodes are used.
@@ -112,6 +120,14 @@ func (env *InteractionEnv) Handle(t *testing.T, d datadriven.TestData) string {
 		//
 		// status 5
 		err = env.handleStatus(t, d)
+	case "tick-election":
+		// Tick an election timeout interval for the given node (but beware the
+		// randomized timeout).
+		//
+		// Example:
+		//
+		// tick-election 3
+		err = env.handleTickElection(t, d)
 	case "tick-heartbeat":
 		// Tick a heartbeat interval.
 		//
--- a/rafttest/interaction_env_handler_set_randomized_election_timeout.go
+++ b/rafttest/interaction_env_handler_set_randomized_election_timeout.go
@@ -1,0 +1,34 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rafttest
+
+import (
+	"testing"
+
+	"github.com/cockroachdb/datadriven"
+	"github.com/stretchr/testify/require"
+)
+
+func (env *InteractionEnv) handleSetRandomizedElectionTimeout(
+	t *testing.T, d datadriven.TestData,
+) error {
+	idx := firstAsNodeIdx(t, d)
+	var timeout int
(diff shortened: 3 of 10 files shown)


testdata/prevote.txt
```diff
@@ -0,0 +1,387 @@
...
+----
+> 1 handling Ready
+  Ready MustSync=false:
+  HardState Term:1 Vote:1 Commit:12
+  CommittedEntries:
+  1/12 EntryNormal "prop_1"
+  Messages:
+  1->2 MsgApp Term:1 Log:1/12 Commit:12
+  1->3 MsgApp Term:1 Log:1/12 Commit:12
+  1->3 MsgPreVoteResp Term:1 Log:0/0 Rejected (Hint: 0)
```
pavelkalinnikov, line 217: Not this PR, just an exploratory question.

I wonder how the `MsgPreVote` sender knows which outbound message this `Resp` corresponds to.

Scenario:
- Node 1 log is outdated. It sends a `MsgPreVote` to node 2.
- Node 1 log is updated, and is now long enough. It sends another `MsgPreVote`.
- Node 2 receives the first message, sends a reject back.
- Node 2 receives the second message, sends an approval back.
- The first (rejected) `Resp` comes back to node 1. Does it consider the first, the second, or both pre-votes rejected?
- If/when the approved `Resp` comes back, how does it identify which pre-vote this corresponds to?

pavelkalinnikov (approved):

pavelkalinnikov:
LGTM, but I can't merge. @tbg @ahrtr Could either of you please?

tbg:
I reviewed this in https://github.com/etcd-io/raft/pull/68. @erikgrinaker could you address my comments (some of which overlap with Pavel's) and fix up this PR? Thanks!

erikgrinaker:
Yep, will do. Thanks both!

testdata/checkquorum.txt
```diff
@@ -0,0 +1,322 @@
...
+  1->2 MsgHeartbeat Term:1 Log:0/0 Commit:11
+  1->3 MsgHeartbeat Term:1 Log:0/0 Commit:11
+  1->2 MsgHeartbeat Term:1 Log:0/0 Commit:11
+  1->3 MsgHeartbeat Term:1 Log:0/0 Commit:11
+  1->2 MsgHeartbeat Term:1 Log:0/0 Commit:11
+  1->3 MsgHeartbeat Term:1 Log:0/0 Commit:11
+  1->2 MsgHeartbeat Term:1 Log:0/0 Commit:11
+  1->3 MsgHeartbeat Term:1 Log:0/0 Commit:11
+  1->2 MsgHeartbeat Term:1 Log:0/0 Commit:11
+  1->3 MsgHeartbeat Term:1 Log:0/0 Commit:11
```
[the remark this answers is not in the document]
erikgrinaker: How so? Conceptually, these heartbeats are sitting in a network buffer somewhere, and we can't know if they've made it to the follower or not. Similarly, inbound messages from the followers can be delayed or dropped as well. Calling stabilize here essentially flushes the network buffer.

testdata/checkquorum.txt
```diff
@@ -0,0 +1,322 @@
+# Tests that CheckQuorum causes a leader to step down if it hasn't heard from a
+# quorum of followers in the past election timeout interval.
+#
+# Also tests that votes are rejected when there is a current leader.
+
+add-nodes 3 voters=(1,2,3) index=10 checkquorum=true
```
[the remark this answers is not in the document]
erikgrinaker: Nice, that's better, thanks!

testdata/prevote.txt
erikgrinaker: I don't think it matters. PreVote is necessarily best-effort, because conditions can chance by the time we hold the actual election. For example, the leader may have woken back up and replicated log entries to the other voters, or a different candidate can also have obtained prevotes and raced with us to win the vote instead. The way to think about it is "if I had held an election right now, would you vote for me?", but it's not a guarantee that we'll actually get the vote, only an indication that we could possibly win an election.

More concretely though, we only record the first vote we receive from a different node:

https://github.com/etcd-io/raft/blob/ffe5efcf2edc7c7236f33eb1f62623c24fc62145/tracker/tracker.go#L261-L264

Once we have a quorum of votes, we'll either hold the election or transition to follower, at which point we clear out our tracked votes:

https://github.com/etcd-io/raft/blob/09c91d8dc0d2f550b69fbe183399e98b1fd6d606/raft.go#L1570-L1584

Also, we won't send multiple `MsgPreVote` in the same precandidate campaign either -- we'll wait for a quorum of prevotes, then transition either to candidate or follower, and then potentially transition back to precandidate and hold another prevote later, at which point we'll send the next `MsgPreVote`. Of course, it's possible that we'll now receive a stale `MsgPreVoteResp` from the previous precandidate campaign if it was delayed while we were transitioning through these states, but such is life.

erikgrinaker:
Ok, comments should be addressed now. Thanks again!

rafttest/interaction_env_handler_set_randomized_election_timeout.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2019 The etcd Authors
```
ahrtr, line 1:
```suggestion
// Copyright 2023 The etcd Authors
```

rafttest/interaction_env_handler.go
```diff
@@ -112,6 +120,13 @@ func (env *InteractionEnv) Handle(t *testing.T, d datadriven.TestData) string {
 		//
 		// status 5
 		err = env.handleStatus(t, d)
+	case "tick-election":
+		// Tick an election timeout interval (but beware the randomized timeout).
```
ahrtr, line 124:
```suggestion
		// Tick an election timeout interval for the given node (but beware
		// the randomized timeout).
```

raft_test.go
```diff
@@ -4978,6 +4978,12 @@ func setRandomizedElectionTimeout(r *raft, v int) {
 	r.randomizedElectionTimeout = v
 }
 
+// SetRandomizedElectionTimeout is like setRandomizedElectionTimeout, but
+// exported for use by tests that are not in the raft package, using RawNode.
+func SetRandomizedElectionTimeout(r *RawNode, v int) {
+	setRandomizedElectionTimeout(r.raft, v)
+}
```
ahrtr, line 4985: Can we just change `setRandomizedElectionTimeout(r *raft, v int)` to `SetRandomizedElectionTimeout(r *raft, v int)`, and  not to add `SetRandomizedElectionTimeout(r *RawNode, v int)`?

ahrtr: OK.

erikgrinaker:
I added another test case for `prevote_checkquorum.txt`: that a quorum of nodes can choose to campaign and win (pre-)votes despite having an active leader, since the PreVote+CheckQuorum recent leader condition only applies for followers.

This is great, because it's exactly the behavior we need in CockroachDB.

ahrtr (approved):
LGTM

Thanks @erikgrinaker

PR #77, opened 2023-06-14
step down when leader removes itself

erikgrinaker:
This patch causes a leader to step down when it removes itself from the group, becoming a follower in the current term. We don't want a removed leader to remain a leader: it no longer accepts proposals, so if it keeps heartbeating then the group is effectively stuck.

cc @tbg @pavelkalinnikov


tbg (approved):

ahrtr:
@erikgrinaker please rebase this PR, I will take a look tomorrow.

erikgrinaker:
TFTR!

PR #78, opened 2023-06-18
add `ForgetLeader`

erikgrinaker:
This patch adds `ForgetLeader()`, which causes a follower to forget its current leader, changing it to None. It remains a leaderless follower in the current term, without campaigning.

This is particularly useful with PreVote+CheckQuorum, if the caller has strong reason to believe the leader is dead, since it will grant prevotes but also revert to follower if it hears from the leader. A quorum of such leaderless followers can thus allow a pre-candidate to hold an election if they believe the leader to be dead.

cc @pavelkalinnikov @tbg 

See https://github.com/cockroachdb/cockroach/pull/105126 for a CockroachDB change that depends on this (there will be a couple more along the same lines).


testdata/forget_leader.txt
```diff
@@ -0,0 +1,423 @@
...
+
+raft-state
+----
+1: StateLeader (Voter) Term:1 Lead:1
+2: StateFollower (Voter) Term:1 Lead:0
+3: StateFollower (Voter) Term:1 Lead:1
+4: StateFollower (Non-Voter) Term:1 Lead:0
+
+# When receiving a heartbeat from the leader, they revert to followers.
+tick-heartbeat 1
```
tbg, line 70: Since you've tested this I assume the answer is no, but in practice, do heartbeats from the (now removed) leader reverse-race with the call to `ForgetLeader`, leadering to a situation where the `ForgetLeader` is wiped out?

testdata/forget_leader.txt
```diff
@@ -0,0 +1,423 @@
...
+ok
+
+raft-state
+----
+1: StateLeader (Voter) Term:1 Lead:1
+2: StateFollower (Voter) Term:1 Lead:1
+3: StateCandidate (Voter) Term:2 Lead:0
+4: StateFollower (Non-Voter) Term:1 Lead:1
+
+stabilize
```
tbg, line 146: Please log-level none this unless there's something relevant in the output here (in which case add comment that explains what to look for please).

testdata/forget_leader.txt
```diff
@@ -0,0 +1,423 @@
...
+INFO 2 forgetting leader 3 at term 2
+
+tick-heartbeat 2
+----
+INFO 2 is starting a new election at term 2
+INFO 2 became candidate at term 3
+INFO 2 [logterm: 2, index: 12] sent MsgVote request to 1 at term 3
+INFO 2 [logterm: 2, index: 12] sent MsgVote request to 3 at term 3
+
+stabilize
```
tbg, line 302: ditto

Might be worth adding a `log-level=X` as a per-command option, i.e.

```
stabilize log-level=none
```

but no worries if you don't want to fry that fish now

tbg (commented):
The code looks good. @ahrtr could you also give this a look? This is adding a new method to the library. It's the result of lengthy experimentation in CRDB - we'd like to avoid waiting for election timeouts when the leader removes itself. This is the best option we've found - in CRDB, we have a designated "next leader" that we campaign when it sees the ConfChange that removes the next leader, but it won't be granted votes if the other replicas still remember the leader. So we call `ForgetLeader` on all followers when they see the ConfChange, which - at least often - allows the election to succeed.

erikgrinaker:
> we'd like to avoid waiting for election timeouts when the leader removes itself

This isn't really the motivating case -- we handle that by calling an immediate election on the leaseholder, bypassing prevote since we know there is no leader to disrupt (https://github.com/cockroachdb/cockroach/pull/104969).

The motivating case is when we unquiesce a range to find a dead leader according to liveness gossip -- if a quorum of replicas consider the leader dead according to liveness we'd like to elect a new leader immediately rather than waiting out the election timeout. We only want to do this when a quorum agrees though, and without disrupting a current leader, since it's possible for individual replicas to have stale or incorrect views of liveness. `ForgetLeader` provides these properties.

testdata/forget_leader.txt
erikgrinaker: Not in the cases we care about in CRDB, no.

For now, we'll be lenient and call this while processing (Pre)Vote requests, under lock, so we'll always grant the (Pre)Vote when appropriate. If we want to be more strict and only do this on unquiescence, then a delayed heartbeat would revert it to follower between a PreVote and Vote, causing it to reject the Vote. But that's ok, and sort of by design since we want to respect a current leader -- in the worst case, it will call another election after waiting out a timeout.

testdata/forget_leader.txt
erikgrinaker: Yep, done.

pavelkalinnikov (approved):
LGTM

node.go
```diff
@@ -575,6 +588,14 @@ func (n *node) TransferLeadership(ctx context.Context, lead, transferee uint64)
 	}
 }
 
+func (n *node) ForgetLeader(ctx context.Context) {
+	select {
+	case n.recvc <- pb.Message{Type: pb.MsgForgetLeader}:
+	case <-n.done:
+	case <-ctx.Done():
+	}
+}
```
ahrtr, line 597:
```suggestion
func (n *node) ForgetLeader(ctx context.Context) error {
	return n.step(ctx, pb.Message{Type: pb.MsgForgetLeader})
}
```

rawnode.go
```diff
@@ -544,6 +544,21 @@ func (rn *RawNode) TransferLeader(transferee uint64) {
...
+// strong reason to believe the leader is dead, since it will grant prevotes but
+// also revert to follower if it hears from the leader. A quorum of such
+// leaderless followers can thus allow a pre-candidate to hold an election if
+// they all believe the leader to be dead.
+//
+// Does nothing with ReadOnlyLeaseBased, since it would allow a new leader to be
+// elected without the old leader knowing.
+func (rn *RawNode) ForgetLeader() {
+	_ = rn.raft.Step(pb.Message{Type: pb.MsgForgetLeader})
+}
```
ahrtr, line 560:
```suggestion
func (rn *RawNode) ForgetLeader() error {
	return rn.raft.Step(pb.Message{Type: pb.MsgForgetLeader})
}
```

node.go
erikgrinaker: Good call, thanks.

rawnode.go
erikgrinaker: 👍

rafttest/interaction_env_handler_stabilize.go
```diff
@@ -25,7 +26,26 @@ import (
 )
 
 func (env *InteractionEnv) handleStabilize(t *testing.T, d datadriven.TestData) error {
-	idxs := nodeIdxs(t, d)
+	idxs := nodeIdxs(t, d) // skips key=value args
+	for _, arg := range d.CmdArgs {
+		for i := range arg.Vals {
+			switch arg.Key {
+			case "log-level":
```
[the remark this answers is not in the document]
erikgrinaker: Yeah, I agree, but don't want to spend time on it now.

rawnode.go
```diff
@@ -544,6 +544,21 @@ func (rn *RawNode) TransferLeader(transferee uint64) {
...
+// remains a leaderless follower in the current term, without campaigning.
+//
+// This is particularly useful with PreVote+CheckQuorum, if the caller has
+// strong reason to believe the leader is dead, since it will grant prevotes but
+// also revert to follower if it hears from the leader. A quorum of such
+// leaderless followers can thus allow a pre-candidate to hold an election if
+// they all believe the leader to be dead.
+//
+// Does nothing with ReadOnlyLeaseBased, since it would allow a new leader to be
+// elected without the old leader knowing.
```
ahrtr, line 557:
```suggestion
// ForgetLeader forgets a follower's current leader, changing it to None.
// See (Node). ForgetLeader for details.
```

erikgrinaker: Done.

rafttest/interaction_env_handler_stabilize.go
```diff
@@ -25,7 +26,26 @@ import (
...
+				var level string
+				arg.Scan(t, i, &level)
+				old := env.Output.Lvl
+				defer func() {
+					env.Output.Lvl = old
+					if level == "none" {
+						env.Output.Builder = &strings.Builder{}
+					}
```
ahrtr, line 41: Could you clarify why do you reset the env.Output.Builder when level == "none"? If we really need to do it, we should get the details encapsulated in `(*InteractionEnv) LogLevel`.
```suggestion
					_ = env.LogLevel(old)
```

erikgrinaker: This is because we not only want to hide the log output, we also want to hide the test output. Consider this example:

```
stabilize
----
> 2 handling Ready
  Ready MustSync=false:
  Lead:0 State:StatePreCandidate
  Messages:
  2->1 MsgPreVote Term:2 Log:1/12
  2->3 MsgPreVote Term:2 Log:1/12
  INFO 2 received MsgPreVoteResp from 2 at term 1
  INFO 2 has received 1 MsgPreVoteResp votes and 0 vote rejections
```

If we only set `env.LogLevel(none)` then it will hide the lines starting with `INFO`, but still output the other lines which are not emitted by the logger but rather the test framework.

This is the same logic as with the `log-level` command, which happens here:

https://github.com/etcd-io/raft/blob/768abf8a1c58ebd9dd859c71b3fba6704400bc82/rafttest/interaction_env_handler.go#L201-L206

erikgrinaker: I went a bit further, and made the `Output.Write*` methods respect the `NONE` log level. This seemed cleaner, and will only suppress output for the duration that `Lvl` was actually set to `NONE`.

One minor effect is that we don't easily know whether we suppressed any output, so we now always output `ok (quiet)` with `NONE`. That seems fine.

node.go
```diff
@@ -189,6 +189,19 @@ type Node interface {
...
+	//
+	// This is particularly useful with PreVote+CheckQuorum, if the caller has
+	// strong reason to believe the leader is dead, since it will grant prevotes
+	// but also revert to follower if it hears from the leader. A quorum of such
+	// leaderless followers can thus allow a pre-candidate to hold an election
+	// if they all believe the leader to be dead.
+	//
+	// Does nothing with ReadOnlyLeaseBased, since it would allow a new leader to
+	// be elected without the old leader knowing.
+	ForgetLeader(ctx context.Context) error
```
ahrtr, line 203: Can we provide an example something like below to make it clearer? 

```
For example, for a three member (e.g 1, 2, and 3) cluster, assuming the current
leader is 3. If we believe that the leader 3 is dead, then we can force member 2
to forget the leader, and make member 1 to campaign. Eventually member 1 will
be elected as a new leader by member 2 and itself. Note if we don't force 
member 2 to forget the leader beforehand, the the campaign from member 1
might be rejected by member 2 because it may think there is already a valid 
leader in current term.
```

ahrtr (approved):

ahrtr:
Overall looks good to me, great work! The only minor concern is the users of the raft library might not be very clear when and how to use `ForgetLeader`. So suggest to add an example in the comment.

node.go
erikgrinaker: How's this?

```
	// ForgetLeader forgets a follower's current leader, changing it to None. It
	// remains a leaderless follower in the current term, without campaigning.
	//
	// This is useful with PreVote+CheckQuorum, where followers will normally not
	// grant pre-votes if they've heard from the leader in the past election
	// timeout interval. Leaderless followers can grant pre-votes immediately, so
	// if a quorum of followers have strong reason to believe the leader is dead
	// (for example via a side-channel or external failure detector) and forget it
	// then they can elect a new leader immediately, without waiting out the
	// election timeout. They will also revert to normal followers if they hear
	// from the leader again, or transition to candidates on an election timeout.
	//
	// For example, consider a three-node cluster where 1 is the leader and 2+3
	// have just received a heartbeat from it. If 2 and 3 believe the leader has
	// now died (maybe they know that an orchestration system shut down 1's VM),
	// we can instruct 2 to forget the leader and 3 to campaign. 2 will then be
	// able to grant 3's pre-vote and elect 3 as leader immediately (normally 2
	// would reject the vote until an election timeout passes because it has heard
	// from the leader recently). However, 3 can not campaign unilaterally, a
	// quorum have to agree that the leader is dead, which avoids disrupting the
	// leader if individual nodes are wrong about it being dead.
	//
	// This does nothing with ReadOnlyLeaseBased, since it would allow a new
	// leader to be elected without the old leader knowing.
```

ahrtr: Looks good to me!

cc @tbg to take another look before we merge this PR later.

rafttest/interaction_env_handler_stabilize.go
```diff
@@ -25,7 +25,22 @@ import (
 )
 
 func (env *InteractionEnv) handleStabilize(t *testing.T, d datadriven.TestData) error {
-	idxs := nodeIdxs(t, d)
+	idxs := nodeIdxs(t, d) // skips key=value args
+	for _, arg := range d.CmdArgs {
```
tbg, line 29: Never `datadriven` has `MaybeScanArgs`, but we haven't cut a release, so this is fine.

tbg (approved):

rafttest/interaction_env_handler_stabilize.go
erikgrinaker: Ok, thanks -- we can deal with that later.

PR #79, opened 2023-06-27
add `StepDownOnRemoval`

erikgrinaker:
In 72d62a1, we made Raft leaders step down when they removed themselves from the group. However, this broke backwards compatibility. This patch makes the behavior opt-in via a new `StepDownOnRemoval` config option, with the intent to enable it unconditionally in the next major release.

In CockroachDB, we campaign a designated voter when the leader is removed from the group. However, during a joint config change, the leader is first demoted to learner before removal. The step-down logic introduced in 72d62a1 triggers on learner demotion (which is the right behavior), but this left the CockroachDB group without a leader, having to wait out an election timeout. While it is straightforward to change this behavior in CockroachDB, we have to maintain backwards compatibility with older nodes during rolling upgrades.

cc @ahrtr @tbg @pavelkalinnikov 

My bad for not testing this change properly in the first place. I'll write up an issue to enable by default in the next major release.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -276,6 +276,13 @@ type Config struct {
 	// This option may be removed once false positives are no longer possible.
 	// See: https://github.com/etcd-io/raft/issues/80
 	DisableConfChangeValidation bool
+
+	// StepDownOnRemoval makes the leader step down when it is removed from the
+	// group or demoted to a learner.
+	//
+	// This behavior will become unconditional in the future. See:
+	// https://github.com/etcd-io/raft/issues/83
+	StepDownOnRemoval bool
 }
 
 func (c *Config) validate() error {
@@ -408,6 +415,7 @@ type raft struct {
 	// when raft changes its state to follower or candidate.
 	randomizedElectionTimeout int
 	disableProposalForwarding bool
+	stepDownOnRemoval         bool
 
 	tick func()
 	step stepFunc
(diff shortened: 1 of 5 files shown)


serathius:
This make me worried about lack of proper backward compatibility testing. Could someone on raft side take an action item to propose a test plan for backward compatibility?

erikgrinaker:
This change is arguably in the gray zone where a bug fix breaks backwards compatibility -- the fact that CockroachDB relies on a learner to commit and replicate the final conf change is problematic. Even so, it's a bug that we currently rely on.

The original PR included test changes for this change in behavior, so in some sense it was already covered by tests (and this PR increases that coverage). The change in behavior was considered beneficial, but in retrospect we should have recognized that any such change in behavior has potential to break something for someone, and should have gated it on a config option in the first place.

tbg (approved):

testdata/confchange_v1_remove_leader.txt
```diff
@@ -176,14 +174,76 @@ stabilize
...
+> 3 handling Ready
+  Ready MustSync=false:
+  Messages:
+  3->1 MsgHeartbeatResp Term:1 Log:0/0
+> 1 receiving messages
+  2->1 MsgHeartbeatResp Term:1 Log:0/0
+  3->1 MsgHeartbeatResp Term:1 Log:0/0
+
+# Just confirming the issue above - leader does not automatically step down.
+raft-state
```
ahrtr, line 245: Expected behaviour: After the electionTimeout, a new leader should can be elected automatically. 

can you please raise a ticket to track this issue?

ahrtr:
> in retrospect we should have recognized that any such change in behavior has potential to break something for someone, and should have gated it on a config option in the first place

Makes sense.

testdata/confchange_v1_remove_leader.txt
erikgrinaker: Opened https://github.com/etcd-io/raft/issues/83. Mind adding it to the 4.0 milestone?

ahrtr: thx & done (added into 4.0 milestone)

ahrtr (approved):
LGTM

thx

Issue #80, opened 2023-07-10
Configuration change validation is unsound


tbg:
We currently validate configuration changes at proposal time

https://github.com/etcd-io/raft/blob/4abd9e927c6d5db930dfdb80237ac584449aeec7/raft.go#L1224-L1260

This is not very helpful because it has false positives (refusing a config change that is actually allowed) though at least it currently doesn't have false negatives (because the set of false positives is sufficiently large 😄)

It could be more effective to either compare against the actual most recent config change in the (including the unstable) log, or to move the verification to apply time (i.e. erroring out conf changes that are illegal).

tbg:
in #81 (which adds a `DisableConfChangeValidation` option), @pavelkalinnikov says:

----

I think we should fix this lag in `raft` more fundamentally, in such a way that the `DisableConfChangeValidation` option is not needed. This mechanism should just not be best-effort, and should not have false positives/negatives. Config changes need some reconsideration.

Relying solely on above-raft config change mechanisms (such as in CRDB) also smells like a footgun. Config changes seem fundamentally belonging to `raft` layer, at least w.r.t. correctness of quorum rules etc.

Currently we do a collection of tricks which makes correctness reasoning hard:
- [scan](https://github.com/etcd-io/raft/blob/4abd9e927c6d5db930dfdb80237ac584449aeec7/raft.go#L916-L919) the log to see if we have an unapplied config
- [guess](https://github.com/etcd-io/raft/blob/4abd9e927c6d5db930dfdb80237ac584449aeec7/raft.go#L887-L892) if we have a [pending](https://github.com/etcd-io/raft/blob/4abd9e927c6d5db930dfdb80237ac584449aeec7/raft.go#L352-L358) config change
- silently [nullify](https://github.com/etcd-io/raft/blob/4abd9e927c6d5db930dfdb80237ac584449aeec7/raft.go#L1256) incorrect config changes, or silently accept them (with the 
`DisableConfChangeValidation` option on).

Instead, I think we should just precisely track and know the current and prospective config from the explicit state, and explicitly error out on incorrect transitions. I.e., when we commit a pending config change, it becomes part of the [state](https://github.com/etcd-io/raft/blob/89d23327cc6bd4ba13650dda3464ee1da37b3898/storage.go#L49-L50). To do so, we would need to decouple config changes from logs (otherwise, again, we would have to have a "scan the log" step to find the next config, e.g. when entries are appended or the commit index is advanced).

[...]

PR #81, opened 2023-07-10
Add Config.DisableConfChangeValidation

tbg:
DisableConfChangeValidation turns off propose-time verification of
configuration changes against the currently active configuration of the
raft instance. These checks are generally sensible (cannot leave a joint
config unless in a joint config, et cetera) but they have false positives
because the active configuration may not be the most recent
configuration. This is because configurations are activated during log
application, and even the leader can trail log application by an
unbounded number of entries.
Symmetrically, the mechanism has false negatives - because the check may
not run against the "actual" config that will be the predecessor of the
newly proposed one, the check may pass but the new config may be invalid
when it is being applied. In other words, the checks are best-effort.

Users should *not* use this option unless they have a reliable mechanism
(above raft) that serializes and verifies configuration changes.

In CockroachDB, we doubly hit the false positive[^1][^2] case and we do have a
higher-level protection mechanism. The intention is for CockroachDB to use this
setting, however it is likely that other systems are susceptible to similar
issues (and perhaps unknowingly so). This setting is thus of general interest.

[^1]: https://github.com/cockroachdb/cockroach/issues/106172#issuecomment-1623568845
[^2]: https://github.com/cockroachdb/cockroach/issues/105797#issuecomment-1613201509

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -254,6 +254,28 @@ type Config struct {
 	// logical clock from assigning the timestamp and then forwarding the data
 	// to the leader.
 	DisableProposalForwarding bool
+
+	// DisableConfChangeValidation turns off propose-time verification of
+	// configuration changes against the currently active configuration of the
+	// raft instance. These checks are generally sensible (cannot leave a joint
+	// config unless in a joint config, et cetera) but they have false positives
+	// because the active configuration may not be the most recent
+	// configuration. This is because configurations are activated during log
+	// application, and even the leader can trail log application by an
+	// unbounded number of entries.
+	// Symmetrically, the mechanism has false negatives - because the check may
+	// not run against the "actual" config that will be the predecessor of the
+	// newly proposed one, the check may pass but the new config may be invalid
+	// when it is being applied. In other words, the checks are best-effort.
+	//
+	// Users should *not* use this option unless they have a reliable mechanism
+	// (above raft) that serializes and verifies configuration changes. If an
+	// invalid configuration change enters the log and gets applied, a panic
+	// will result.
+	//
+	// This option may be removed once false positives are no longer possible.
+	// See: https://github.com/etcd-io/raft/issues/80
+	DisableConfChangeValidation bool
 }
 
 func (c *Config) validate() error {
@@ -356,6 +378,9 @@ type raft struct {
 	// be proposed if the leader's applied index is greater than this
 	// value.
 	pendingConfIndex uint64
+	// disableConfChangeValidation is Config.DisableConfChangeValidation,
+	// see there for details.
+	disableConfChangeValidation bool
 	// an estimate of the size of the uncommitted tail of the Raft log. Used to
 	// prevent unbounded log growth. Only maintained by the leader. Reset on
 	// term changes.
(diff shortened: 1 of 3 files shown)


pavelkalinnikov (approved):
LGTM, with some thoughts on config changes.

raft.go
```diff
@@ -254,6 +254,23 @@ type Config struct {
...
 	DisableProposalForwarding bool
+
+	// DisableConfChangeValidation turns off propose-time verification of
+	// configuration changes against the currently active configuration of the
+	// raft instance. These checks are generally sensible (cannot leave a joint
+	// config unless in a joint config, et cetera) but they have false positives
+	// because the active configuration may not be the most recent
+	// configuration. This is because configurations are activated during log
+	// application, and even the leader can trail log application by an
+	// unbounded number of entries.
```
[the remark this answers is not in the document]
tbg: High level agree - but this is intentionally the tactical fix. Apologies, I neglected to link this issue to the PR description which I had filed to have that discussion: https://github.com/etcd-io/raft/pull/81

I will copy your comment there.

raft.go
```diff
@@ -254,6 +254,23 @@ type Config struct {
...
+	// Symmetrically, the mechanism has false negatives - because the check may
+	// not run against the "actual" config that will be the predecessor of the
+	// newly proposed one, the check may pass but the new config may be invalid
+	// when it is being applied. In other words, the checks are best-effort.
+	//
+	// Users should *not* use this option unless they have a reliable mechanism
+	// (above raft) that serializes and verifies configuration changes.
+	DisableConfChangeValidation bool
```
ahrtr, line 273: I think the exposed configuration item is confusing to users.  If there are any issues on the conf change validation or conf change itself, then we should only fix the raft side instead of delegating part of the responsibility/fix to upper applications. Providing more flexibility here (exposing the conf item) also provides a dangerous hole for users to make mistake (e.g. turning on the item: setting it to `true` wrongly).

Usually users (upper applications) completely depends on raft to reach an agreement on the conf change. They shouldn't worry about anything on conf change at all. 

Can you just workaround this issue on application side for now, and In the meanwhile, reconsider how to improve the conf change in raft in the long-term?

tbg: > Can you just workaround this issue on application side for now, and In the meanwhile, reconsider how to improve the conf change in raft in the long-term?

That was what I tried at first, unfortunately it was not successful and convinced me that it is not the right approach here. The invariant we want in CockroachDB is that if you make a proposal and hand it to raft often enough, it will eventually come up for application (where CockroachDB determines whether it will have an effect or apply as a no-op, subject to our replay protection mechanism, lease system, etc). This check breaks this invariant, and since the check is unstable (produces different results at different times) there is the scenario in which we cannot tell whether the proposal made it into the log, which means we can't release certain locks until we see it in application. It's a bit complicated to explain, but essentially working around it in CRDB is something I will rule out on account of the undue complexity it imposes.

> If there are any issues on the conf change validation or conf change itself, then we should only fix the raft side instead of delegating part of the responsibility/fix to upper applications

I would describe what I am doing here as a tactical fix. The incorrect check in raft is a problem, and I agree with you that it should be addressed (which is tracked in https://github.com/etcd-io/raft/issues/80). However, I expect that to be a nontrivial change, and there should be a way to opt out of the problematic behavior until then - which is what this patch achieves.

> Usually users (upper applications) completely depends on raft to reach an agreement on the conf change. They shouldn't worry about anything on conf change at all.

This might be true in most simple systems, but it is not true in CockroachDB. We currently avoid drift between the state machine's tracked config and the raft config, and we do so by updating the tracked config through raft conf changes. So it's not acceptable to us for raft to drop configuration changes that would be valid. And at the same time, it means that we have necessarily already validated the configuration change before it happens and establish mutual exclusion of subsequent conf changes.
I can't rule out that some of these boundary conditions will change in the future, but I don't see it as an anti pattern regardless. The raft library should be useful for advanced use as well as simple use cases.

ahrtr: thx for the clarification. At the minimum, I think we need to clearly add a comment to mention that `DisableConfChangeValidation` is just a temporary solution for https://github.com/etcd-io/raft/issues/80, may be removed after the issue has been resolved in raft.

tbg: Good point - added the comment. PTAL.

ahrtr (approved):

Issue #73, opened 2023-06-05
RFC: Message level mechanism for disabling proposal forwarding


mitake:
The Raft library provides the parameter [DisableProposalForwarding](https://github.com/etcd-io/raft/blob/main/raft.go#L256). When the flag is true, it doesn’t forward MsgProp type messages from follower to leader. I guess providing a similar option for each `Message` might be valuable. It’s great if I can get comments on this idea.

Context: a program which uses the Raft library can have asynchronous processes which can issue Raft request from a server process. In the case of etcd, [lease and compaction](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/server.go#L772) are typical examples. I guess other users of the library might have similar processes.

Such an asynchronous process can be implemented a goroutine whose behavior depends a condition that its node is leader or not. If the node is a leader, the goroutine issues Raft messages asynchronously (and periodically).

This behavior might be problematic in some cases. If the goroutine can be paused by various reasons (high load on CPU, disk I/O, etc), the node can be a follower by a new leader election. The problem is that the goroutine can behave based on a stale information that the node is still a leader. In such a case, the goroutine can issue Raft requests because it thinks that it’s still a leader. If the messages shouldn’t be duplicated, it might be harmful. Otherwise it will be problematic. In the case of etcd, it can cause lease revoking from a stale leader: https://github.com/etcd-io/etcd/issues/15247

Note that this problem should happen only if the Raft library logic recognizes itself as a follower:
- If the Raft library logic recognizes itself as a leader: `MsgProp` messages will be handled by the stale node itself and result `MsgApp`. In this case other nodes can reject the message because these messages have an old term.
- If the Raft library logic recognizes itself as a follower already: `MsgProp` will be forwarded to a new leader and the new leader will send `MsgApp`. Although the original source of the messages is the stale leader, the new cluster can accept the messages.

(the above behavior is quite subtle and I'm still checking it, I'm glad if I can get feedback on it too)

I guess setting `DisableProposalForwarding` true will be a simple solution for avoiding this situation. However, if a program which uses the Raft library doesn’t provide a client side mechanism of selecting a leader and sending messages to it (e.g. etcd clientv3), the parameter will make the program not functional because it affects all messages. So I think it’s nice if the Raft library can provide a mechanism to disable proposal forwarding only for specific message types.

There might be some possible approaches:
- Adding a new flag to `Message`: it’s simple but be too large change.
[...]

PR #88, opened 2023-07-24
Add Config.DisableProposalForwardingCallback for controlling message level proposal forwarding

mitake:
This PR adds the new field `DisableProposalForwardingCallback` to the Config struct. The field is a pointer to callback function which takes `raftpb.Message`. A follower node calls this callback when a state machine tries to submit a `MsgProp` message. If the callback returns true based on the message, the follower node drops it and returns `ErrProposalDropped`.

If we pass a callback function which always returns true, the behavior of the Raft library will be same to the case of `DisableProposalForwarding` == true.

The motivation is described in https://github.com/etcd-io/raft/issues/73


ahrtr:
@mitake 
- How will you define the callback in etcd?
- Providing such flexibility might be a little dangerous,  because the raft's behavior/correction depends on application's code logic.

mitake:
I checked [the dissertation](https://github.com/ongardie/dissertation) and section 6.2 `Routing requests to the leader` has descriptions related to this topic:
```
1. The first option, which we recommend and which LogCabin implements, is for the server to reject the 
request and return to the client the address of the leader, if known. This allows the client to reconnect to 
the leader directly, so future requests can proceed at full speed. It also takes very little additional code to 
implement, since clients already need to reconnect to a different server in the event of a leader failure.

2. Alternatively, the server can proxy the client’s request to the leader. This may be simpler in some cases. 
For example, if a client connects to any server for read requests (see Section 6.4), then proxying the 
client’s write requests would save the client from having to manage a dis- tinct connection to the leader 
used only for writes.
```
According to this description, MsgProp (corresponding to 2) is an optional mechanism for Raft. So `DisableProposalForwarding` can be provided by the library and it would be safe to have additional mechanism for dropping specific proposal messages. If clientv3 can have a mechanism to select a leader in client side like what clientv2 does: https://github.com/etcd-io/etcd/pull/4030 it might be possible to turn on `DisableProposalForwarding` in etcd (not fully sure though)? 

```
Leaders: A server might be in the leader state, but if it isn’t the current leader, it could be needlessly 
delaying client requests. For example, suppose a leader is partitioned from the rest of the cluster, but it 
can still communicate with a particular client. Without additional mechanism, it could delay a request 
from that client forever, being unable to replicate a log entry to any other servers. Meanwhile, there 
might be another leader of a newer term that is able to communicate with a majority of the cluster and 
would be able to commit the client’s request. Thus, a leader in Raft steps down if an election timeout 
elapses without a successful round of heartbeats to a majority of its cluster; this allows clients to retry 
their requests with another server.
```
I think this description is related to our lease revoking problem by a stale leader. In our case a client is etcd server itself, so a delayed lease revoke message will be forwarded to a leader and accepted.

Anyway I'll try this change with the failpoint based testing when I have time.

mitake:
Related idea: I feel it might be good for controlling risks of new parameter of the Raft library by adding a new struct (e.g. `Experimental`) to `Config`, which collects such new parameters. Library users can know that parameters in `Experimental` are experimental thing and can be deprecated in future.

I think we have a few candidate of such parameters like https://github.com/etcd-io/etcd/pull/7782 
cc @chaochn47 you might be interested in this idea? It's great if I can have your opinion.

ahrtr:
A couple of points:
- In network isolation status, a node may still consider itself as a leader but actually it might not be. It will still broadcast the leaseRevoke request, and eventually the request will be sent to other nodes when the network is recovered.
- This solution can't resolve the second case "Leases are wrongly revoked by the new leader" as mentioned in the Root Cause section in [doc](https://docs.google.com/document/d/1peLDjwebnSuNR69ombuUiJ5fkSw7Rd3Slv_mUpOLsmg/edit)

ahrtr:
> * In network isolation status, a node may still consider itself as a leader but actually it might not be. It will still broadcast the leaseRevoke request, and eventually the request will be sent to other nodes when the network is recovered.

It seems not a problem, because it will be rejected by other nodes due to it's smaller term.

mitake:
Sorry let me check the entire of your doc later and reply in a few days @ahrtr . Let me reply to the below comment:

> It seems not a problem, because it will be rejected by other nodes due to it's smaller term.

For MsgApp of LeaseRevoke, this is true. But before that MsgProp of LeaseRevoke is sent from the stale leader (etcd layer thinks itself as a leader but raft layer isn't), and it cannot be rejected because MsgProp doesn't have term information and it will be just forwarded to a new leader. I think this behavior is quite complicated so would like to write a doc including diagrams.

PR #96, opened 2023-08-18
Optimise memory deallocation for stable log entries

### LINKED ISSUE #71: Release in-memory log space based on entries size
When entries exit the `unstable` log structure https://github.com/etcd-io/raft/blob/3e6cb625f59ca9da34d03df484217d874340e794/log_unstable.go#L156

it shrinks itself based on the entries slice length:
https://github.com/etcd-io/raft/blob/3e6cb625f59ca9da34d03df484217d874340e794/log_unstable.go#L162-L179

Firstly, the `len(u.entries)*lenMultiple < cap(u.entries)` does not necessarily do a right thing. After shifting the slice start in `u.entries = u.entries[num:]`, the first `num` entries "disappear" from `len` and `cap` of this slice: https://go.dev/play/p/ao1xaL0Edla.

So it's possible that `len >= cap/2` all the time, even though only a small portion of the backing slice is used. We should take the `cap` of the entire backing slice into account instead.

Secondly, this heuristic only takes `len/cap` of the slice into consideration, but not the size of the entries referenced by it. It could be beneficial to, in addition, shrink the slice if the sum size of  the "garbage" entries if more than a half. This would keep the overall memory consumption more controlled.

Aditya-Sood:
Increment: Modify unstable.shrinkEntriesArray() to better account for underlying-array utilisation
Fixes #71


Aditya-Sood:
hi @pavelkalinnikov, could you please review the PR?
if it looks good then I'll get started on the "[MemoryStorage.Compact](https://github.com/etcd-io/raft/blob/3e6cb625f59ca9da34d03df484217d874340e794/storage.go#L251) method" as well that you had mentioned in the original issue

Aditya-Sood:
hi @pavelkalinnikov, does the PR need any changes?

Aditya-Sood:
hi @pavelkalinnikov, could you please review the PR?

pavelkalinnikov:
@Aditya-Sood Sorry, been super busy recently. Will review this PR this week.

log_unstable.go
```diff
@@ -152,30 +152,52 @@ func (u *unstable) stableTo(i, t uint64) {
...
 // entries wouldn't be safe because clients might still be using them.
-func (u *unstable) shrinkEntriesArray() {
-	// We replace the array if we're using less than half of the space in
-	// it. This number is fairly arbitrary, chosen as an attempt to balance
-	// memory usage vs number of allocations. It could probably be improved
-	// with some focused tuning.
-	const lenMultiple = 2
-	if len(u.entries) == 0 {
+func (u *unstable) shrinkEntriesSlice(nextUnstableEntryIndex int) {
+	if nextUnstableEntryIndex == len(u.entries) { //all log entries have been successfully stored
```
pavelkalinnikov, line 166: nit: Please add a whitespace between `//` and the comment. Same bellow.

log_unstable.go
```diff
@@ -152,30 +152,52 @@ func (u *unstable) stableTo(i, t uint64) {
...
+func (u *unstable) doEntriesNeedCompaction(nextUnstableEntryIndex int) bool {
+	//Eligible for compaction if the stable entries occupy more than (1/lenMultiple) times of the:
+	// a) underlying-array indexes, OR
+	// b) total memory occupied by entries (both stable and unstable)
+	const lenMultiple = 2
+
+	countStableEntries := nextUnstableEntryIndex
+	if countStableEntries > cap(u.entries)/lenMultiple {
+		return true
 	}
```
pavelkalinnikov, line 184: This heuristic is also different from the proposed idea in #71. It still honours only "visible" entries in deciding when to compact. The idea was that we track the allocated slice size, and decide based on that.

It also seems different from the original: it checks the ratio of stable entries vs. cap; previously it checked the ratio of all entries vs. cap. Any particular reason why you went with this heuristic?

pavelkalinnikov (commented):
Thanks for the PR. Posting some initial thoughts.

This change needs unit-tests to check correctness and demonstrate the situations in which the compaction heuristic kicks in.

log_unstable.go
Aditya-Sood: hi @pavelkalinnikov, thanks for the review
I forgot that `cap()` considers the backing array only for the last index, not the beginning one

I had another thought - at the time of shrinking the `entries` slice, if we remove references to the shrunk `pb.Entry` elements, then two things will happen:
1. with no more references, the 'shrunk' `pb.Entry` values will immediately become available for GC (if not in use)
2. the `append()` method will consider creating an entirely new backing array at the time of adding new elements when capacity is low, and then the preceding unreachable values (which currently hold empty `pb.Entry` values after the above change) will also become available for GC

this seemed like a much cleaner solution to me

if this approach seems feasible, then we can optimise the design further by replacing `pb.Entry` values with pointers instead, to further minimise the memory cost of holding shrunk values in the array until `append()` replaces the backing array

Aditya-Sood: hi @pavelkalinnikov, could you please share your review of the above approach?

Aditya-Sood:
hi @pavelkalinnikov, could you please share your thoughts on this approach - https://github.com/etcd-io/raft/pull/96#discussion_r1327965688

Aditya-Sood:
hi @ahrtr, would you be able to weigh-in on https://github.com/etcd-io/raft/pull/96#discussion_r1327965688?

Aditya-Sood:
hi @mitake, @serathius, @ptabor, @spzala
would you be able to share your opinion on https://github.com/etcd-io/raft/pull/96#discussion_r1327965688?
I can start working on the change then, this PR has been open for 2.5 months now

Aditya-Sood:
hi @pav-kv

serathius:
QQ for any optimisation, do we have a benchmark that confirms the improvement?

Aditya-Sood:
hi @serathius, no I haven't implemented the proposed change yet
if you don't foresee any immediate issues with this approach then I can implement it and look into benchmarking?

PR #105, opened 2023-09-30
Return the `actually committed index` if it's greater than the `official committed index` for `ReadIndex`

ahrtr:
When etcd receives committed Entries, it does the following two things concurrently,
- apply the entries and respond to client
- persist the hard state

It's possible that it has successfully responded to the client, but fails to persist the hardstate, such as crashes right before persisting the hardstate(including committed index).  After etcd starts again, 
- The initial committed index of the raft is an out of date value;
- When a client sends a linearizable request, it gets the readIndex firstly, and get the old value; accordingly won't wait for the  new entries being applied, and respond the client with an stale data.
   - Note that etcd replays WAL entries on bootstrap, but this step may be executed before the replaying finishes.


raft.go
```diff
@@ -1303,7 +1303,10 @@ func stepLeader(r *raft, m pb.Message) error {
 	case pb.MsgReadIndex:
 		// only one voting member (the leader) in the cluster
 		if r.prs.IsSingleton() {
-			if resp := r.responseToReadIndexReq(m, r.raftLog.committed); resp.To != None {
+			// It's possible that the "officially committed index" < "actually committed index",
+			// raft may return an out of date readindex in such case.
+			mci := r.prs.Committed()
+			if resp := r.responseToReadIndexReq(m, max(r.raftLog.committed, mci)); resp.To != None {
```
serathius, line 1309:
```suggestion
			readIndex := max(r.raftLog.committed, r.prs.Committed())
			if resp := r.responseToReadIndexReq(m, readIndex); resp.To != None {
```

ahrtr: updated

raft.go
```diff
@@ -1303,7 +1303,10 @@ func stepLeader(r *raft, m pb.Message) error {
 	case pb.MsgReadIndex:
 		// only one voting member (the leader) in the cluster
 		if r.prs.IsSingleton() {
-			if resp := r.responseToReadIndexReq(m, r.raftLog.committed); resp.To != None {
+			// It's possible that the "officially committed index" < "actually committed index",
```
serathius, line 1306: I think I understand what you mean, however the naming "officially commited index" vs "actually commited index" don't make sense for me. Commited index is just a commited index. I think should have a clearer explanation about difference between r.raftLog.commited vs r.prs.Commited()

serathius:
Not yet convinced that the problem is with readIndex. I would be interested more into persistence of HardState.
HardState should be also persisted to WAL, isn't this guaranteed to be written?

ahrtr:
See https://github.com/etcd-io/etcd/pull/16675.

This PR is a standalone change for raft.

raft.go
ahrtr: see the comment for method `readIndex`

ahrtr:
Won't continue to work on this. 

https://github.com/etcd-io/etcd/issues/16666#issuecomment-1747077487

Issue #130, opened 2024-01-22
flow control: append messages


pav-kv:
This issue describes examples of limited control of the data flows in `raft` package, and what this can lead to. This motivates improvements in this area. The umbrella issue for this is #64.

Consider a general class of systems[^1] for which a single server can host multiple/many instances of `raft` [Node](https://github.com/etcd-io/raft/blob/e22adc041041d1980ddefd974c4ea414edd90cd4/node.go#L131-L132) / [RawNode](https://github.com/etcd-io/raft/blob/e22adc041041d1980ddefd974c4ea414edd90cd4/rawnode.go#L31-L34). Say, in a cluster of O(N) servers each server is hosting O(K) instances of `raft.RawNode`. K can be in [tens of] thousands. `RawNode`s of a single `raft` group (typically, 3x or 5x replicas per group) are distributed across the cluster arbitrarily, e.g. one group can be on servers `{0, 1, 2}`, another group on servers `{0, 10, 57}`, etc.

In the picture below, we first consider the `MsgApp` data flows from `raft` leaders on server 0 to followers on other N-1 servers.

![fan-out](https://github.com/etcd-io/raft/assets/3757441/b7e49547-4579-480f-b127-d03bd144a66f)

In the current design of `raft`, each `RawNode` acts independently of others, and makes local decisions on when to [send](https://github.com/etcd-io/raft/blob/main/raft.go#L595-L600) a `MsgApp` to a follower, based on merely the fact that the follower is behind on the log. The `RawNode` does this as soon as possible (at any `Step/Ready` iteration), and the application/server layer has little control of this behaviour. The `RawNode` drives the process, and the application has to adapt.

Such design is prone to a bunch of overload scenarios.

### Scenario 1: Leader-side overload

<details>
<summary>Details</summary>

In a loaded system, followers are always behind by a little, and sometimes by much (e.g. after a server crash/restart). As a result, many `RawNode`s always independently decide to send a `MsgApp` to their followers. Each `RawNode` fetches a few entries, constructs a `MsgApp`, and sends to the application layer via the `Ready` [struct](https://github.com/etcd-io/raft/blob/e22adc041041d1980ddefd974c4ea414edd90cd4/node.go#L49-L115).

Server 0 has limited capacity. If there are too many messages, server 0 can crash because of reaching an out-of-memory situation.

To prevent this, the application layer has a few options, which can be categorised as “workarounds”:

[...]

Issue #131, opened 2024-01-24
redundant MsgApp heartbeats


pav-kv:
#### Background

Append messages [[MsgApp](https://github.com/etcd-io/raft/blob/4fcf99f38c20868477e01f5f5c68ef1e4377a8b1/raftpb/raft.proto#L88-L89)] that leader [sends](https://github.com/etcd-io/raft/blob/4fcf99f38c20868477e01f5f5c68ef1e4377a8b1/raft.go#L660-L663) to a follower mainly serve 3 purposes:

1. Update the follower's log. The message contains a batch of entries extending the log.
2. Update the "commit index" for a follower whose commit index might be behind.
3. Learn the follower's log state.

Upon a successful `MsgApp` processing, the follower sends a `MsgAppResp` back to the leader, [confirming](https://github.com/etcd-io/raft/blob/4fcf99f38c20868477e01f5f5c68ef1e4377a8b1/raft.go#L1739) the new size of its log. **Note that it does not confirm the update of the commit index.**

#### Problem

The follower can be completely up-to-date (both the log and the commit index), but the leader can still send empty `MsgApp` for purpose (2), to update the commit index (e.g. [here](https://github.com/etcd-io/raft/blob/4fcf99f38c20868477e01f5f5c68ef1e4377a8b1/raft.go#L1503-L1505)). The conditions for sending these empty messages are brittle and probabilistic.

It would be good to have a simpler condition for sending commit index update messages, such that it is hard to get wrong and doesn't require guessing for the follower.

#### Solution

Observe that it is only necessary to send a commit index update to a follower if the follower's commit index is behind. This is because the commit index never regresses (it is [stored](https://github.com/etcd-io/raft/blob/4fcf99f38c20868477e01f5f5c68ef1e4377a8b1/raftpb/raft.proto#L110-L113) in `HardState`).

The conditions for sending an empty `MsgApp` with a commit index update will become simpler if the leader tracks the follower's commit index (similarly to how it [tracks](https://github.com/etcd-io/raft/blob/e22adc041041d1980ddefd974c4ea414edd90cd4/tracker/progress.go#L31) the follower's log size for deciding when to send new log entries).

To support this, the follower should notify the leader of the `Commit` index. It should be as simple as setting the `Commit` field of the `MsgAppResp` message that it [sends](https://github.com/etcd-io/raft/blob/4fcf99f38c20868477e01f5f5c68ef1e4377a8b1/raft.go#L1739) to the leader.

[...]

PR #134, opened 2024-01-25
raft: consolidate all append message sending

pav-kv:
This PR consolidates all decision-making about sending append messages into a single `maybeSendAppend` method. Previously, the behaviour depended on the `sendIfEmpty` flag which was set/unset depending on the circumstances in which the method is called. This is unnecessary because the `Progress` struct contains enough information about the leader->follower flow state, so `maybeSendAppend` can be made stand-alone.

In a follow-up PR, the consolidated `maybeSendAppend` method will be used to implement a more flexible message flow control.

Part of #130

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1563,7 +1563,7 @@ func stepLeader(r *raft, m pb.Message) error {
 		// If snapshot finish, wait for the MsgAppResp from the remote node before sending
 		// out the next MsgApp.
 		// If snapshot failure, wait for a heartbeat interval before next try
-		pr.MsgAppFlowPaused = true
+		pr.PauseMsgAppProbes(true)
 	case pb.MsgUnreachable:
 		// During optimistic replication, if the remote becomes unreachable,
 		// there is huge probability that a MsgApp is lost.
--- a/raft.go
+++ b/raft.go
@@ -588,24 +588,24 @@ func (r *raft) send(m pb.Message) {
 	}
 }
 
-// sendAppend sends an append RPC with new entries (if any) and the
-// current commit index to the given peer.
-func (r *raft) sendAppend(to uint64) {
-	r.maybeSendAppend(to, true)
-}
-
-// maybeSendAppend sends an append RPC with new entries to the given peer,
-// if necessary. Returns true if a message was sent. The sendIfEmpty
-// argument controls whether messages with no entries will be sent
-// ("empty" messages are useful to convey updated Commit indexes, but
-// are undesirable when we're sending multiple messages in a batch).
+// maybeSendAppend sends an append RPC with log entries (if any) that are not
+// yet known to be replicated in the given peer's log, as well as the current
+// commit index. Usually it sends a MsgApp message, but in some cases (e.g. the
+// log has been compacted) it can send a MsgSnap.
 //
-// TODO(pav-kv): make invocation of maybeSendAppend stateless. The Progress
-// struct contains all the state necessary for deciding whether to send a
-// message.
-func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
+// In some cases, the MsgApp message can have zero entries, and yet being sent.
+// When the follower log is not fully up-to-date, we must send a MsgApp
+// periodically so that eventually the flow is either accepted or rejected. Not
+// doing so can result in replication stall, in cases when a MsgApp is dropped.
+//
+// Returns true if a message was sent, or false otherwise. A message is not sent
+// if the follower log and commit index are up-to-date, the flow is paused (for
+// reasons like in-flight limits), or the message could not be constructed.
+func (r *raft) maybeSendAppend(to uint64) bool {
 	pr := r.trk.Progress[to]
-	if pr.IsPaused() {
+
+	last, commit := r.raftLog.lastIndex(), r.raftLog.committed
+	if !pr.ShouldSendMsgApp(last, commit) {
 		return false
 	}
 
@@ -617,35 +617,25 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 		return r.maybeSendSnapshot(to, pr)
 	}
 
-	var ents []pb.Entry
-	// In a throttled StateReplicate only send empty MsgApp, to ensure progress.
-	// Otherwise, if we had a full Inflights and all inflight messages were in
-	// fact dropped, replication to that follower would stall. Instead, an empty
-	// MsgApp will eventually reach the follower (heartbeats responses prompt the
-	// leader to send an append), allowing it to be acked or rejected, both of
-	// which will clear out Inflights.
-	if pr.State != tracker.StateReplicate || !pr.Inflights.Full() {
-		ents, err = r.raftLog.entries(pr.Next, r.maxMsgSize)
-	}
-	if len(ents) == 0 && !sendIfEmpty {
-		return false
-	}
-	// TODO(pav-kv): move this check up to where err is returned.
-	if err != nil { // send a snapshot if we failed to get the entries
-		return r.maybeSendSnapshot(to, pr)
+	var entries []pb.Entry
+	if pr.CanSendEntries(last) {
+		if entries, err = r.raftLog.entries(pr.Next, r.maxMsgSize); err != nil {
+			// Send a snapshot if we failed to get the entries.
+			return r.maybeSendSnapshot(to, pr)
+		}
 	}
 
-	// Send the actual MsgApp otherwise, and update the progress accordingly.
+	// Send the MsgApp, and update the progress accordingly.
 	r.send(pb.Message{
 		To:      to,
 		Type:    pb.MsgApp,
 		Index:   prevIndex,
 		LogTerm: prevTerm,
-		Entries: ents,
-		Commit:  r.raftLog.committed,
+		Entries: entries,
+		Commit:  commit,
 	})
-	pr.SentEntries(len(ents), uint64(payloadsSize(ents)))
-	pr.SentCommit(r.raftLog.committed)
+	pr.SentEntries(len(entries), uint64(payloadsSize(entries)))
+	pr.SentCommit(commit)
 	return true
 }
 
@@ -704,7 +694,7 @@ func (r *raft) bcastAppend() {
 		if id == r.id {
 			return
 		}
-		r.sendAppend(id)
+		r.maybeSendAppend(id)
 	})
 }
 
@@ -1482,7 +1472,7 @@ func stepLeader(r *raft, m pb.Message) error {
 				if pr.State == tracker.StateReplicate {
 					pr.BecomeProbe()
 				}
-				r.sendAppend(m.From)
+				r.maybeSendAppend(m.From)
 			}
 		} else {
 			// We want to update our tracking if the response updates our
@@ -1521,21 +1511,13 @@ func stepLeader(r *raft, m pb.Message) error {
 					// to respond to pending read index requests
 					releasePendingReadIndexMessages(r)
 					r.bcastAppend()
-				} else if r.id != m.From && pr.CanBumpCommit(r.raftLog.committed) {
-					// This node may be missing the latest commit index, so send it.
-					// NB: this is not strictly necessary because the periodic heartbeat
-					// messages deliver commit indices too. However, a message sent now
-					// may arrive earlier than the next heartbeat fires.
-					r.sendAppend(m.From)
 				}
-				// We've updated flow control information above, which may
-				// allow us to send multiple (size-limited) in-flight messages
-				// at once (such as when transitioning from probe to
-				// replicate, or when freeTo() covers multiple messages). If
-				// we have more entries to send, send as many messages as we
-				// can (without sending empty messages for the commit index)
+				// We've updated flow control information above, which may allow us to
+				// send multiple (size-limited) in-flight messages at once (such as when
+				// transitioning from probe to replicate, or when freeTo() covers
+				// multiple messages). Send as many messages as we can.
 				if r.id != m.From {
-					for r.maybeSendAppend(m.From, false /* sendIfEmpty */) {
+					for r.maybeSendAppend(m.From) {
 					}
 				}
 				// Transfer leadership is in progress.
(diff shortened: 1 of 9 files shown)


raft.go
```diff
@@ -588,54 +588,60 @@ func (r *raft) send(m pb.Message) {
...
+// This may send an empty append message (with no entries) if replication to
+// this follower is throttled, or there are no new entries but the commit index
+// for the follower can be bumped.
+func (r *raft) sendAppend(to uint64, pr *tracker.Progress) bool {
+	if pr.State == tracker.StateProbe {
+		return !pr.MsgAppFlowPaused && r.maybeSendAppend(to, pr)
+	} else if pr.State != tracker.StateReplicate {
 		return false
-	}
+	} // only StateReplicate below
```
nvanbenschoten, line 602: Do you think it's worth being exhaustive here, as a means to be more explicit:
```go
switch pr.State {
case tracker.StateProbe:
    return !pr.MsgAppFlowPaused && r.maybeSendAppend(to, pr)
case tracker.StateSnapshot:
    return false
case StateReplicate:
    // Fall through.
default:
    panic("unreachable")
}
```

raft.go
```diff
@@ -588,54 +588,60 @@ func (r *raft) send(m pb.Message) {
...
+	if pr.CanBumpCommit(r.raftLog.committed) {
+		return r.maybeSendEmptyAppend(to, pr)
+	}
+	// In a throttled StateReplicate, send an empty append message if we haven't
+	// done so recently.
+	//
+	// We must send periodic appends so that eventually the follower either
+	// accepts or rejects it. If we don't do so, replication can stall if all the
+	// in-flight appends are lost/dropped.
+	return !pr.MsgAppFlowPaused && pr.Match < r.raftLog.lastIndex() &&
```
nvanbenschoten, line 622: nit: this would read better if you split out the two different conditions for sending an empty message from the logic to do so. Something like:
```go
// NB: the commit index is periodically sent in the heartbeat messages, so
// technically we don't need the CanBumpCommit clause here to guarantee commit
// index convergence on the follower. However, sending it via MsgApp here
// allows faster (no heartbeat interval delay) convergence in some cases.
sendEmpty := pr.CanBumpCommit(r.raftLog.committed)

// In a throttled StateReplicate, send an empty append message if we haven't
// done so recently.
//
// We must send periodic appends so that eventually the follower either
// accepts or rejects it. If we don't do so, replication can stall if all the
// in-flight appends are lost/dropped. 
sendEmpty = sendEmpty || !pr.MsgAppFlowPaused && pr.Match < r.raftLog.lastIndex()
	
if sendEmpty {
	return r.maybeSendEmptyAppend(to, pr)
}
return false
```
Up to you though.

raft.go
```diff
@@ -588,54 +588,60 @@ func (r *raft) send(m pb.Message) {
...
+		r.maybeSendEmptyAppend(to, pr)
+}
 
+// maybeSendAppend sends a non-empty append message to the given follower. It
+// may send a snapshot instead if the required section of the log is no longer
+// available in this leader's log. Returns true if a message was sent.
```
nvanbenschoten, line 628: Could you mention when it returns false in the comment? It currently reads as if false indicates that a snapshot was sent.

raft.go
```diff
@@ -649,6 +655,29 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
...
+		// The log probably got truncated at >= pr.Next, so we can't catch up the
+		// follower log anymore. Send a snapshot instead.
+		return r.maybeSendSnapshot(to, pr)
+	}
+	r.send(pb.Message{
+		To:      to,
+		Type:    pb.MsgApp,
+		Index:   pr.Next - 1,
+		LogTerm: prevTerm,
+		Commit:  r.raftLog.committed,
```
nvanbenschoten, line 674: I'm noticing the odd asymmetry between `MsgApp` sending the latest Commit index and the follower computing `min(committed, lastnewi)`, and `MsgHeartbeat` carefully sending `min(pr.Match, r.raftLog.committed)` to avoid a follower-side panic. Do you have thoughts about that?

tracker/progress.go
```diff
@@ -161,9 +162,7 @@ func (pr *Progress) SentEntries(entries int, bytes uint64) {
 			pr.Next += uint64(entries)
 			pr.Inflights.Add(pr.Next-1, bytes)
 		}
-		// If this message overflows the in-flights tracker, or it was already full,
-		// consider this message being a probe, so that the flow is paused.
-		pr.MsgAppFlowPaused = pr.Inflights.Full()
+		pr.MsgAppFlowPaused = true
```
nvanbenschoten, line 165: Is this commit changing the meaning of `MsgAppFlowPaused`?

raft.go
pav-kv: Yes. The commit index in `MsgApp` is "optimistic", and can be larger than `pr.Match`. As a result, a follower can simultaneously append and commit entries in some cases. Technically, `Commit` index can even exceed the last index of the entries in the batch. The way I think about `MsgApp` is that it's effectively 2 independent messages:
1. A [logSlice](https://github.com/etcd-io/raft/blob/83d8deca057fef78cf3a86ea70457e4d3255e4da/types.go#L38-L67) of the leader log to append to the follower log.
2. A `(term, commit)` is the point in the leader@`term` log that is committed (see #144 explaining the "leader log" coordinate system).

As such, even if the append (1) is redundant/no-op (because the follower already has these entries from elsewhere), the commit index (2) can still communicate valuable information - if the follower log is at the same term, it can advance the commit index safely. We don't take advantage of this currently (the commit index is [advanced](https://github.com/etcd-io/raft/blob/83d8deca057fef78cf3a86ea70457e4d3255e4da/log.go#L127) only within the bounds of the appended entries (1)), but we should.

The commit index in `MsgHeartbeat` is "pessimistic" and cut at `pr.Match` because currently the follower has a [panic](https://github.com/etcd-io/raft/blob/83d8deca057fef78cf3a86ea70457e4d3255e4da/raft.go#L1802) if the index is not in the bounds of the follower log. This panic is unnecessary though, see #138 and #139 aiming to relax it.

I would like to clean-up the semantics of the `Commit` indices in both messages, to both be treated as (2) above - a `(term, index)` that the leader sends without relying on the knowledge about the follower log. The follower log has enough knowledge to apply it safely.

tracker/progress.go
pav-kv: Yes. It wasn't very well specified before.

Now `MsgAppFlowPaused` means the flow of **empty** "probe" `MsgApp` messages is paused, if it's ongoing. But if the `MsgApp` flow is not throttled, this flag is ignored/no-op (see `sendAppend`), though still can be `true` or `false`.

I think we should unexport this flag, it's internal and changed its meaning a couple of times (last time in https://github.com/etcd-io/etcd/pull/14633). We don't use it in CRDB. @ahrtr Does etcd use it?

tracker/progress.go
```diff
@@ -169,6 +176,19 @@ func (pr *Progress) SentEntries(entries int, bytes uint64) {
...
+// CanBumpCommit returns true if sending the given commit index can potentially
+// advance the follower's commit index.
+func (pr *Progress) CanBumpCommit(index uint64) bool {
+	return pr.pendingCommit < min(index, pr.Next-1)
+}
+
+// SentCommit updates the pendingCommit.
+func (pr *Progress) SentCommit(commit uint64) {
+	if commit > pr.pendingCommit {
+		pr.pendingCommit = min(commit, pr.Next-1)
```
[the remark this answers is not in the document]
pav-kv: There is no strict requirement that `pendingCommit` does not regresses. The `pendingCommit` means there is an in-flight message with this commit index. But since there is no guarantee that this message is delivered, we may need to send this commit again.

`pendingCommit` is the equivalent of `Next`, but for commit indices. Both may regress for same reasons (when we have no acks from the follower). I chose to maintain `pendingCommit < Next` invariant. Given this invariant, we have:

`pr.pendingCommit < commit` and `pr.pendingCommit < pr.Next`. This line bumps the `pendingCommit` but no higher than `pr.Next - 1`, to maintain the invariant.

Should probably add some commenting here.

tracker/progress.go
pav-kv: We do [use](https://github.com/cockroachdb/cockroach/blob/254dbd247fb8ed352a11439063b29f23a0767f28/pkg/kv/kvserver/replica_raft_overload.go#L121-L123) the `IsPaused()` helper though, which uses this field. The dependency is not trivial here. Maybe `IsPaused()` should be kept up-to-date with the decision-making in `sendAppend`, i.e. `IsPaused() == true` iff `sendAppend` does not send anything.

ahrtr: > @ahrtr Does etcd use it?

No, etcd doesn't use it.

tracker/progress.go
```diff
@@ -248,14 +255,47 @@ func (pr *Progress) IsPaused() bool {
...
+// In StateReplicate, generally a message is sent if there are log entries that
+// are not yet in-flight, and the in-flight limits are not exceeded. Otherwise,
+// we don't send a message, or send a "probe" message in a few situations.
+//
+// A probe message (containing no log entries) is sent if the follower's commit
+// index can be updated, or there hasn't been a probe message recently. We must
+// send a message periodically even if all log entries are in-flight, in order
+// to guarantee that eventually the flow is either accepted or rejected.
+//
+// In StateReplicate, we do not send append messages.
```
pav-kv, line 289: StateSnapshot

raft.go
```diff
@@ -588,24 +588,19 @@ func (r *raft) send(m pb.Message) {
 	}
 }
 
-// sendAppend sends an append RPC with new entries (if any) and the
+// maybeSendAppend sends an append RPC with new entries (if any) and the
```
pav-kv, line 591: TODO: clean-up this method comment.

raft.go
pav-kv: The PR has evolved a bit, the is no longer relevant.

tracker/progress.go
pav-kv: I reworked the PR in such a way that `MsgAppFlowPaused` can now be unexported (left a TODO for that). The behaviour of `IsPaused()` method hasn't change, although I think it should be removed in the future.

raft.go
pav-kv: Made this comment more detailed.

pav-kv:
@ahrtr PTAL. Also asking @bdarnell to review (@nvanbenschoten and @erikgrinaker can't look at at it this week).

ahrtr:
Will take a closer look this week. 

A generic comment, in order to make the review easier, suggest to breakdown this PR. Such as, 
- get code refactor, method/function renaming, comment updating etc. in one PR, so that they can be reviewed and merged asap. 
- If you change any logic or adding any new field, then it needs super careful review, please try to make such PR as small as possible (e.g. < 100 lines of code change).

pav-kv:
@bdarnell We have an issue for the commit index acks you described: #131. And a draft PR for it: #132. It's not necessary for the flow control change here though.

tracker/progress.go
```diff
@@ -126,13 +131,15 @@ func (pr *Progress) BecomeProbe() {
 	} else {
 		pr.ResetState(StateProbe)
 		pr.Next = pr.Match + 1
+		pr.pendingCommit = min(pr.pendingCommit, pr.Match)
```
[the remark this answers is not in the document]
ahrtr: Suggest to always use `pr.Next - 1` to keep it conceptually correct and easy to understand. 

This comment also applies to the change in `(*Progress) MaybeDecrTo`

tracker/progress.go
```diff
@@ -163,20 +166,25 @@ func (pr *Progress) BecomeSnapshot(snapshoti uint64) {
...
-		pr.MsgAppFlowPaused = pr.Inflights.Full()
-	case StateProbe:
-		pr.MsgAppFlowPaused = true
-	default:
-		panic(fmt.Sprintf("sending append in unhandled state %s", pr.State))
+	if pr.State == StateReplicate && entries > 0 {
+		pr.Next += uint64(entries)
+		pr.Inflights.Add(pr.Next-1, bytes)
 	}
+	pr.PauseMsgAppProbes(true)
```
ahrtr, line 173: It seems that it changes the original logic. In previous implementation, if `pr.State == StateReplicate && !pr.Inflights.Full()` is `true`, then it will not pause the flow.  Your new implementation will pause the flow in this case. Is it an intentional change?

tracker/progress.go
```diff
@@ -226,19 +258,57 @@ func (pr *Progress) MaybeDecrTo(rejected, matchHint uint64) bool {
...
+// to guarantee that eventually the flow is either accepted or rejected.
+//
+// In StateSnapshot, we do not send append messages.
+func (pr *Progress) ShouldSendMsgApp(last, commit uint64) bool {
+	switch pr.State {
+	case StateProbe:
+		return !pr.MsgAppProbesPaused
+	case StateReplicate:
+		return pr.CanBumpCommit(commit) ||
+			pr.Match < last && (!pr.MsgAppProbesPaused || pr.CanSendEntries(last))
```
ahrtr, line 304: The condition is a little over complicated. Is it possible that both `pr.MsgAppProbesPaused` and `pr.CanSendEntries(last)` are `true`? `ShouldSendMsgApp` may still return `true` in such case?

raft.go
```diff
@@ -601,9 +601,7 @@ func (r *raft) send(m pb.Message) {
 // Returns true if a message was sent, or false otherwise. A message is not sent
 // if the follower log and commit index are up-to-date, the flow is paused (for
 // reasons like in-flight limits), or the message could not be constructed.
-func (r *raft) maybeSendAppend(to uint64) bool {
-	pr := r.trk.Progress[to]
-
+func (r *raft) maybeSendAppend(to uint64, pr *tracker.Progress) bool {
```
ahrtr, line 604: What's the intention to pass the `tracker.Progress`? Save a couple of cpu cycles? If yes, then I think readability outweighs the performance here. The lookup in map should be very fast.

tracker/progress.go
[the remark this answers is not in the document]
pav-kv: Done in #171.

tracker/progress.go
pav-kv: Yes, it's possible. `MsgAppProbesPaused == true` means: if we are in the probing/throttled mode, don't send a message. But if we are not in this mode (`CanSendEntries` is true) then `MsgAppProbesPaused` is effectively ignored.

Keeping `MsgAppProbesPaused` in sync with `CanSendEntries` would require some additional invariants.

raft.go
pav-kv: Yes, just "normalizing" the code here. All the callers of `maybeSendAppend` already have a `Progress`. Will remove this commit though.

pav-kv:
Closing this PR, since it has been superseded by https://github.com/cockroachdb/cockroach/pull/124006 in CRDB.

PR #135, opened 2024-01-25
remove unused min and max funcs

pav-kv:
Go has min/max built-ins since 1.21.
https://tip.golang.org/doc/go1.21#language


ahrtr (approved):
lgtm

PR #137, opened 2024-01-25
testdata: add a lagging commit test

pav-kv:
This commit adds a test demonstrating the effect of delayed commit on a follower node after a network hiccup between the leader and this follower.

In the described scenario, after the moment of committing an entry on the leader, it takes `HeartbeatInterval + 3/2 * RTT` until the follower learns this entry is committed.

This is suboptimal, and could take `HeartbeatInverval + 1/2 * RTT` if the leader didn't cut the commit index at `Progress.Match` before sending it to the follower.

Informs #138

### DIFF
--- a/testdata/lagging_commit.txt
+++ b/testdata/lagging_commit.txt
@@ -1,0 +1,174 @@
+# This test demonstrates the effect of delayed commit on a follower node after a
+# network hiccup between the leader and this follower.
+
+# Skip logging the boilerplate. Set up a raft group of 3 nodes, and elect node 1
+# as the leader. Nodes 2 and 3 are the followers.
+log-level none
+----
+ok
+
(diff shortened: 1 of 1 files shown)


pav-kv:
@nvanbenschoten @ahrtr PTAL.

erikgrinaker:
>This is suboptimal, and could take HeartbeatInverval + 1/2 * RTT if the leader didn't cut the commit index at Progress.Match before sending it to the follower.

Why do we care? Follower reads?

pav-kv:
> Why do we care? Follower reads?

Yes, and also the general "convergence" of the system. For other kinds of reads, like quorum reads, this also seems relevant.

pav-kv:
@ahrtr Good to merge?

Issue #144, opened 2024-01-30
The notion of Term is overloaded


pav-kv:
#### Problem

Currently, a `raft` node has a `Term` [field](https://github.com/etcd-io/raft/blob/026484c68b7bc59cd208e216d35e1ff747d24d9c/raft.go#L341) (and its storage [counterpart](https://github.com/etcd-io/raft/blob/026484c68b7bc59cd208e216d35e1ff747d24d9c/raftpb/raft.proto#L111) in `HardState`) which is the latest term seen by this node.

- This term may or may not have a vote from this/other nodes, and may or may not have a leader.
- The local `raftLog`, correspondingly, may or may not be consistent with the leader's log at this term, and the leader's log might not exist in the first place.

This uncertainty is a source of workarounds and potential safety bugs, and is generally confusing. To combat this, let's try to bring some structure, and then improve the codebase.

TODO: add action items.
TODO: add code links.
TODO: keep writing, this is a WIP text.
TODO: convert this to a tech note / design doc.

#### Background

Every leader in `raft` has a unique term. During election, a prospective leader (candidate) negotiates a starting point `(prevIndex, prevTerm)`, and then appends a contiguous run of entries at index `prevIndex+1` and higher. The set of all entries in the system thus forms a tree. Raft implements a bunch of safety rules that govern how the `(prevIndex, prevTerm)` point is chosen (it has to be in the subtree of the "committed" point in this tree), and when entries are considered committed (a quorum of replicas are in the same subtree).

Consider an example of such a "tree" representing the log evolution:

```
	term
	▲
	│
	│               ┌→5 5 5
	│               │
	│	        ┌→4 4
	│           │
	│     ┌→3 3 3 3 3 3 3 3 3 3 3 3
	│     │
	│     │ ┌→2 2 2 2 2 2
	│     │ │
	│   1 1 1 1 1 1
	└──────────────────────────────────────────► index
```

Every node in the `raft` group is at some "point" in this tree. A node's log is the unique path in this tree from `(0, 0)` to entry `(term, index)` at the "tip" of the log.

The "leader log" is the path from `(0, 0)` to the latest entry that this leader appended. For the picture above, here are the "leader logs" corresponding to the leader term:

```
	t1: 1 1 1 1 1 1
	t2: 1 1 1 2 2 2 2 2 2
	t3: 1 1 3 3 3 3 3 3 3 3 3 3 3 3
	t4: 1 1 3 3 3 4 4
[...]

Issue #142, opened 2024-01-27
raftLog: decouple log data structure and flow control


pav-kv:
The [raftLog](https://github.com/etcd-io/raft/blob/026484c68b7bc59cd208e216d35e1ff747d24d9c/log.go#L24) structure currently plays multiple roles. Notably:

- Provides read/write [access](https://github.com/etcd-io/raft/blob/026484c68b7bc59cd208e216d35e1ff747d24d9c/log.go#L25-L34) to the raft log (both the `unstable` in-memory part, and the `Storage` part), and keeps the basic metadata about the log.
- Implements [flow control](https://github.com/etcd-io/raft/blob/026484c68b7bc59cd208e216d35e1ff747d24d9c/log.go#L35-L62) mechanisms for applying the commands from this log.

These are 2 distinct roles, and should be decoupled. Once decoupling is done, `raftLog` can be isolated in its own package and rigorously tested without assumptions about the flow.

Elbehery:
@pav-kv  can i take this ?

pav-kv:
Sure, give it a try. I got to say that `good first issue` might be an understatement, doing this requires some familiarity with this repo. LMK if I can help figuring things out.

The complexity may vary here depending on how far we go with it. But we can start with simple mechanical decoupling, and revisiting the tests.

Also, let's not merge big changes before the release #89 is done. We can prepare the change in parallel though.

Elbehery:
sgtm, please assign me 👍🏽

pav-kv:
@Elbehery Feel free to send PRs / contribute and tag things "Part of #142". This is more like an umbrella issue doesn't need an assignee, IMO. I'm also working on related parts at the moment.

nizanshami:
Is this still needed? This looks like a nice task, and I would be happy to work on it.

ramses:
> Is this still needed? This looks like a nice task, and I would be happy to work on it.

+1

samuelbartels20:
@pav-kv Please has this issue been resolved. If not, i would like love to work on it

pav-kv:
I don't think it's resolved, feel free to contribute / send PRs. Please note that I'm not active on this repository, you might want to coordinate with maintainers/approvers.

IAMRogerXi:
is anyone still actively working on this one? if not, i will start a new branch and work on it.....

IAMRogerXi:
i have set up the dev env and run the existing unit testing to understand the basic flow. will try to start the design soon.

[...]

PR #145, opened 2024-01-30
raft: add entryID and logSlice types

pav-kv:
This PR introduces type-safe wrappers for log entry IDs and raft log slices. We will use them for a more readable and safe code. Some usage is introduced in this PR, see individual commits.

A lot of code in this repo manipulates `(term, index)` tuples as separate raw ints. Likewise, entry slices are handled without association with the leader term. Getting this usage wrong can be costly, since both entry IDs and log slices are centrepiece of raft safety.

Related to: #142, #144

### DIFF
--- a/log.go
+++ b/log.go
@@ -107,7 +107,7 @@ func (l *raftLog) String() string {
 // maybeAppend returns (0, false) if the entries cannot be appended. Otherwise,
 // it returns (last index of new entries, true).
 func (l *raftLog) maybeAppend(prev entryID, committed uint64, ents ...pb.Entry) (lastnewi uint64, ok bool) {
-	if !l.matchTerm(prev.index, prev.term) {
+	if !l.matchTerm(prev) {
 		return 0, false
 	}
 
@@ -150,13 +150,15 @@ func (l *raftLog) append(ents ...pb.Entry) uint64 {
 // a different term.
 // The index of the given entries MUST be continuously increasing.
 func (l *raftLog) findConflict(ents []pb.Entry) uint64 {
-	for _, ne := range ents {
-		if !l.matchTerm(ne.Index, ne.Term) {
-			if ne.Index <= l.lastIndex() {
+	for i := range ents {
+		if id := pbEntryID(&ents[i]); !l.matchTerm(id) {
+			if id.index <= l.lastIndex() {
+				// TODO(pav-kv): can simply print %+v of the id. This will change the
+				// log format though.
 				l.logger.Infof("found conflict at index %d [existing term: %d, conflicting term: %d]",
-					ne.Index, l.zeroTermOnOutOfBounds(l.term(ne.Index)), ne.Term)
+					id.index, l.zeroTermOnOutOfBounds(l.term(id.index)), id.term)
 			}
-			return ne.Index
+			return id.index
 		}
 	}
 	return 0
@@ -436,12 +438,12 @@ func (l *raftLog) isUpToDate(lasti, term uint64) bool {
 	return term > l.lastTerm() || (term == l.lastTerm() && lasti >= l.lastIndex())
 }
 
-func (l *raftLog) matchTerm(i, term uint64) bool {
-	t, err := l.term(i)
+func (l *raftLog) matchTerm(id entryID) bool {
+	t, err := l.term(id.index)
 	if err != nil {
 		return false
 	}
-	return t == term
+	return t == id.term
 }
 
 func (l *raftLog) maybeCommit(maxIndex, term uint64) bool {
--- a/log.go
+++ b/log.go
@@ -106,23 +106,25 @@ func (l *raftLog) String() string {
 
 // maybeAppend returns (0, false) if the entries cannot be appended. Otherwise,
 // it returns (last index of new entries, true).
-func (l *raftLog) maybeAppend(prev entryID, committed uint64, ents ...pb.Entry) (lastnewi uint64, ok bool) {
-	if !l.matchTerm(prev) {
+func (l *raftLog) maybeAppend(a logSlice, committed uint64) (lastnewi uint64, ok bool) {
+	if !l.matchTerm(a.prev) {
 		return 0, false
 	}
+	// TODO(pav-kv): propagate logSlice down the stack. It will be used all the
+	// way down in unstable, for safety checks, and for useful bookkeeping.
 
-	lastnewi = prev.index + uint64(len(ents))
-	ci := l.findConflict(ents)
+	lastnewi = a.prev.index + uint64(len(a.entries))
+	ci := l.findConflict(a.entries)
 	switch {
 	case ci == 0:
 	case ci <= l.committed:
 		l.logger.Panicf("entry %d conflict with committed entry [committed(%d)]", ci, l.committed)
 	default:
-		offset := prev.index + 1
-		if ci-offset > uint64(len(ents)) {
-			l.logger.Panicf("index, %d, is out of range [%d]", ci-offset, len(ents))
+		offset := a.prev.index + 1
+		if ci-offset > uint64(len(a.entries)) {
+			l.logger.Panicf("index, %d, is out of range [%d]", ci-offset, len(a.entries))
 		}
-		l.append(ents[ci-offset:]...)
+		l.append(a.entries[ci-offset:]...)
 	}
 	l.commitTo(min(committed, lastnewi))
 	return lastnewi, true
--- a/log.go
+++ b/log.go
@@ -450,12 +450,12 @@ func (l *raftLog) matchTerm(id entryID) bool {
 	return t == id.term
 }
 
-func (l *raftLog) maybeCommit(maxIndex, term uint64) bool {
-	// NB: term should never be 0 on a commit because the leader campaigns at
-	// least at term 1. But if it is 0 for some reason, we don't want to consider
-	// this a term match in case zeroTermOnOutOfBounds returns 0.
-	if maxIndex > l.committed && term != 0 && l.zeroTermOnOutOfBounds(l.term(maxIndex)) == term {
-		l.commitTo(maxIndex)
+func (l *raftLog) maybeCommit(at entryID) bool {
+	// NB: term should never be 0 on a commit because the leader campaigned at
+	// least at term 1. But if it is 0 for some reason, we don't consider this a
+	// term match.
+	if at.term != 0 && at.index > l.committed && l.matchTerm(at) {
+		l.commitTo(at.index)
 		return true
 	}
 	return false
--- a/log.go
+++ b/log.go
@@ -372,12 +372,14 @@ func (l *raftLog) stableSnapTo(i uint64) { l.unstable.stableSnapTo(i) }
 // to Ready().
 func (l *raftLog) acceptUnstable() { l.unstable.acceptInProgress() }
 
-func (l *raftLog) lastTerm() uint64 {
-	t, err := l.term(l.lastIndex())
+// lastEntryID returns the ID of the last entry in the log.
+func (l *raftLog) lastEntryID() entryID {
+	index := l.lastIndex()
+	t, err := l.term(index)
 	if err != nil {
-		l.logger.Panicf("unexpected error when getting the last term (%v)", err)
+		l.logger.Panicf("unexpected error when getting the last term at %d: %v", index, err)
 	}
-	return t
+	return entryID{term: t, index: index}
 }
 
 func (l *raftLog) term(i uint64) (uint64, error) {
@@ -435,7 +437,8 @@ func (l *raftLog) allEntries() []pb.Entry {
 // whichever log has the larger lastIndex is more up-to-date. If the logs are
 // the same, the given log is up-to-date.
 func (l *raftLog) isUpToDate(lasti, term uint64) bool {
-	return term > l.lastTerm() || (term == l.lastTerm() && lasti >= l.lastIndex())
+	last := l.lastEntryID()
+	return term > last.term || term == last.term && lasti >= last.index
 }
 
 func (l *raftLog) matchTerm(id entryID) bool {
--- a/log.go
+++ b/log.go
@@ -430,15 +430,16 @@ func (l *raftLog) allEntries() []pb.Entry {
 	panic(err)
 }
 
-// isUpToDate determines if the given (lastIndex,term) log is more up-to-date
+// isUpToDate determines if a log with the given last entry is more up-to-date
 // by comparing the index and term of the last entries in the existing logs.
+//
 // If the logs have last entries with different terms, then the log with the
 // later term is more up-to-date. If the logs end with the same term, then
 // whichever log has the larger lastIndex is more up-to-date. If the logs are
 // the same, the given log is up-to-date.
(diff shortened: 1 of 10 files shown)


pav-kv:
@ahrtr @nvanbenschoten This addresses one of the feedback points from #139 review.

PTAL. Commit-by-commit review might be the most convenient.

types.go
```diff
@@ -0,0 +1,50 @@
...
+	index uint64
+}
+
+// logAppend is the metadata that describes a validated raft log append.
+//
+// Every log append is done on behalf of a leader at a specific term. This term
+// does not necessarily match entryID.term of the entries being appended, since
+// a leader can send any entries from earlier terms in its own log.
+//
+// TODO(pav-kv): add entries, and helpers for validating the invariants of the append,
```
pav-kv, line 33: After we add `entries` to this, we can rename `logAppend` to a more generic `logSlice`. For **any** slice of the log, all these invariants are true, and all the fields are useful.

Specifically, the `unstable` log is a `logSlice` in which `prev` is the snapshot entry ID. The in-storage log is the physical counterpart of `logSlice`.

We can then consolidate most of the append arithmetics in `unstable` into this struct. This makes it a nice unit, that we can test properly without needing the boilerplate for setting up `unstable`/`raftLog` etc.

types.go
```diff
@@ -0,0 +1,50 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package raft
+
+// entryID uniquely identifies a raft log entry.
+//
+// Every entry is associated with a leadership term which issued this entry and
+// initially appended it to the log. There can only be one leader at any term,
+// and a leader never issues entries with the same index.
```
pav-kv, line 21: s/never issues entries/never issues two entries/

types.go
```diff
@@ -0,0 +1,50 @@
...
+type entryID struct {
+	term  uint64
+	index uint64
```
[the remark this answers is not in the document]
nvanbenschoten: I think what you have here is better, so that a reader just looking at the struct definition can understand.

raft.go
```diff
@@ -1730,16 +1730,19 @@ func stepFollower(r *raft, m pb.Message) error {
...
 		return
 	}
-	if mlastIndex, ok := r.raftLog.maybeAppend(m.Index, m.LogTerm, m.Commit, m.Entries...); ok {
+	if mlastIndex, ok := r.raftLog.maybeAppend(a, m.Commit, m.Entries...); ok {
 		r.send(pb.Message{To: m.From, Type: pb.MsgAppResp, Index: mlastIndex})
 		return
 	}
 	r.logger.Debugf("%x [logterm: %d, index: %d] rejected MsgApp [logterm: %d, index: %d] from %x",
-		r.id, r.raftLog.zeroTermOnOutOfBounds(r.raftLog.term(m.Index)), m.Index, m.LogTerm, m.Index, m.From)
+		r.id, r.raftLog.zeroTermOnOutOfBounds(r.raftLog.term(a.prev.index)), a.prev.index, a.prev.term, a.prev.index, m.From)
```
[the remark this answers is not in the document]
pav-kv: Yeah, I guess that's true. I was hoping to reduce the printing boilerplate - all these index/term pairs can be printed as a single unit. But it's a good point that at this level we should be printing the message input.

types.go
```diff
@@ -0,0 +1,50 @@
...
+
+package raft
+
+// entryID uniquely identifies a raft log entry.
+//
+// Every entry is associated with a leadership term which issued this entry and
+// initially appended it to the log. There can only be one leader at any term,
+// and a leader never issues entries with the same index.
+type entryID struct {
+	term  uint64
```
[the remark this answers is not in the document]
pav-kv: This will potentially blow up the size of this PR. Let's keep this TODO in mind.

pav-kv:
@nvanbenschoten Addressed your comments, thank you. PTAL.

types.go
```diff
@@ -0,0 +1,50 @@
...
+	index uint64
+}
+
+// logAppend is the metadata that describes a validated raft log append.
```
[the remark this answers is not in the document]
pav-kv: I.e. the semantics of these appends are "from leader to log".
When this is generalized to a `logSlice`, the semantics will be "a sub-slice of the leader log at `logSlice.term`".

types.go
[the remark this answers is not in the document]
pav-kv: Added a TODO in comment.

types.go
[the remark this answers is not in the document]
pav-kv: The leader appends happen [here](https://github.com/pav-kv/raft/blob/d14e61bd82582490766aef1d4b13986fa2573a53/raft.go#L797-L813). That `raftLog.append` method should probably be [unified with](https://github.com/pav-kv/raft/blob/main/log.go#L107-L140) `maybeAppend` into one.

For a leader appending to its own log, the safety check on `Term` would trivially be true.

types.go
```diff
@@ -0,0 +1,52 @@
...
+//
+// TODO(pav-kv): consider having type aliases for term and index.
+type entryID struct {
+	term  uint64
+	index uint64
+}
+
+// logAppend is the metadata that describes a validated raft log append.
+//
+// Every log append is done on behalf of a leader at a specific term. This term
```
pav-kv, line 31: `s/a leader/an elected leader/`

This is a crucial bit that distinguishes `raft.Term` from the `term` in this struct. The `term` in this struct necessarily means that there is a leader at this term.

pav-kv:
@nvanbenschoten @ahrtr Things start stacking up on top of this PR. Please let me know if there is major objection to this clean-up. If not, I would like to merge and iterate/rebase.

log.go
```diff
@@ -452,12 +452,12 @@ func (l *raftLog) matchTerm(id entryID) bool {
...
 }
 
-func (l *raftLog) maybeCommit(maxIndex, term uint64) bool {
-	// NB: term should never be 0 on a commit because the leader campaigns at
-	// least at term 1. But if it is 0 for some reason, we don't want to consider
-	// this a term match in case zeroTermOnOutOfBounds returns 0.
-	if maxIndex > l.committed && term != 0 && l.zeroTermOnOutOfBounds(l.term(maxIndex)) == term {
-		l.commitTo(maxIndex)
+func (l *raftLog) maybeCommit(at entryID) bool {
+	// NB: term should never be 0 on a commit because the leader campaigned at
```
pav-kv, line 456: This `term != 0` check is probably unnecessary, but leaving as is in case there is some corner case that doesn't handle this properly. There should be an invariant that well protects us here.

nvanbenschoten (approved):
Looks great!

pav-kv:
Thanks!
@ahrtr Could you take a final pass?

types.go
pav-kv: Done in this PR.

types.go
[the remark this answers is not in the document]
pav-kv: Generalized the `logAppend` struct to be `logSlice` in this PR. It will be a central part of a few more clean-ups, so it's good lay the right foundation.

pav-kv:
@nvanbenschoten I've changed this PR a little since the last review. Mainly, I reworked the `logSlice` struct, and improved commenting for it. Otherwise the change is equivalent.

@ahrtr @serathius Could you please take a final look at this PR, and merge if it looks good? Commit-by-commit review can be most convenient. I have follow-up work that propagates safe struct usage in other places, but I am not including them in this PR to keep it small and simple. Thank you.

types.go
```diff
@@ -0,0 +1,108 @@
...
+
+// valid returns nil iff the logSlice is a well-formed log slice. See logSlice
+// comment for details on what constitutes a valid raft log slice.
+func (a logSlice) valid() error {
+	prev := a.prev
+	for i := range a.entries {
+		id := pbEntryID(&a.entries[i])
+		if id.term < prev.term || id.index != prev.index+1 {
+			return fmt.Errorf("entries %+v and %+v not consistent", prev, id)
+		}
```
ahrtr, line 101:
```suggestion
		if id.term < prev.term || id.term > a.term || id.index != prev.index+1 {
			return fmt.Errorf("entries %+v and %+v not consistent, leader term: %d", prev, id, a.term)
		}
```

types.go
```diff
@@ -0,0 +1,108 @@
...
+// or in a test code that manually hard-codes this struct. In these cases, the
+// invariants should be validated using the valid() method.
+type logSlice struct {
+	// term is the leader term containing the given entries in its log.
+	term uint64
+	// prev is the ID of the entry immediately preceding the entries.
+	prev entryID
+	// entries contains the consecutive entries representing this slice.
+	entries []pb.Entry
+}
```
ahrtr, line 76: I am thinking it might be better to replace the following fields in Message with this struct.

https://github.com/etcd-io/raft/blob/2f109970c68e505c223f4d3d5c13a8dcd8f7d528/raftpb/raft.pb.go#L402-L415

This can be addressed in separate PRs. In order to be backward compatible, we might want to do it across multiple releases. e.g. add logSlice into protobuf firstly, and then replace the fields in Message in a following release.

types.go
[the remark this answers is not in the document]
ahrtr: I tend to keep it as it's. It adds unnecessary complexity (type cast) and accordingly reduce the readability.  The naming `term` and `index` are already good enough to self-describe the meaning.

ahrtr:
A generic comment, I prefer not to add too much small commits, as most of them are mechanical changes, e.g. `use xxx in xxx`. So suggest to squash the commits, or get the `type.go and types_test.go` in a commit, and get all others in another commit.

types.go
```diff
@@ -0,0 +1,108 @@
...
+//
+// Property (1) means the slice is contiguous. Properties (2) and (3) mean that
+// the terms of the entries in a log never regress. Property (4) means that a
+// leader log at a specific term never has entries from higher terms.
+//
+// Users of this struct can assume the invariants hold true. Exception is the
+// "gateway" code that initially constructs logSlice, such as when its content
+// is sourced from a message that was received via transport, or from Storage,
+// or in a test code that manually hard-codes this struct. In these cases, the
+// invariants should be validated using the valid() method.
```
ahrtr, line 68: Can you explain more about the exceptions? I think we should add a generic verification to validate the invariant properties in each test if possible.

types.go
```diff
@@ -0,0 +1,108 @@
...
+// lastEntryID returns the ID of the last entry in this log slice, or prev if
+// there are no entries.
+func (a logSlice) lastEntryID() entryID {
+	if ln := len(a.entries); ln != 0 {
+		return pbEntryID(&a.entries[ln-1])
+	}
+	return a.prev
+}
+
+// valid returns nil iff the logSlice is a well-formed log slice. See logSlice
```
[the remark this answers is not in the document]
pav-kv: "iff" stands for "[if and only if](https://en.wikipedia.org/wiki/If_and_only_if)", and communicates the semantics more accurately than "if"

types.go
pav-kv: We don't need to compare every entry `term` with `a.term`. The last entry has the biggest term, so it's enough to only check the last one (see a couple of lines below).

types.go
pav-kv: Yeah, we should probably leave this for some future consideration. The scope of this work currently is internal cleanup that doesn't affect public API / wire protocol.

There are a few additional considerations:

- `logSlice` is intended as a code-level data structure. It's usually a bad practice to use protobuf-generated structs as data structures because we have limited control of its content, and also this is part of the public API.
- The `LogTerm/Index` fields are used by a bunch of other message types (not only `MsgApp`). We can't easily deprecate them, rather it would have to be a new `LogSlice` field or something. A broader refactoring could be: split the `Message` into message-type-specific sub-structs.

There is some argument for moving in a slightly opposite direction from protobufs:
- `raft` doesn't need to know every `pb.Entry`, it only internally cares about config changes. Exposing application-specific entry contents to this package poses security and privacy risks. There could be a world in which `raft` is purely a code-level algorithm/protocol, and messages are packaged at application level.
- `raft` necessitates dependency on protobuf, which is not the most efficient format: it has to be marshaled/unmarshaled. Some applications may choose other wire formats, for instance `flatbuffers`.

pav-kv:
@ahrtr
> A generic comment, I prefer not to add too much small commits, as most of them are mechanical changes

I prefer to isolate mechanical changes to separate commits, so that all changes have a small scope and are obviously correct. Such a PR "tells a story" rather than dumps a large change onto a reviewer. Small changes make it possible to review one thing at a time, and not have to do mental work to decompose it. I do this both for the reviewer and for myself, to self-review and convince myself too.

I usually communicate in the PR that commit-by-commit may be convenient / increase confidence in this being a mechanical change. This is not mandatory to review things this way though - if you prefer the other way around feel free to review the whole change rather than individual commits.

types.go
pav-kv: Agree with that. This PR adds a `valid()` check to `TestLogMaybeAppend`, which is at the moment the only user of the new struct. More tests will be converted in a follow-up PR. The `valid()` check in other tests helped to find a few places where these invariants are violated (so the tests are testing incorrect behaviour).

types.go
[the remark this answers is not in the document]
pav-kv: Removed the TODO.

types.go
pav-kv: Modified the error message to contain the leader term though, this would be useful for debugging.

types.go
pav-kv: The "exception" is basically any place which receives a log slice from an untrusted source (message, storage, etc). Example: see the `handleAppendEntries` call: https://github.com/etcd-io/raft/blob/2c8980be8c0d4a9aded784913a1ec5e07f3c1cce/raft.go#L1748-L1750

Once the invariants are validated, this struct can be passed around internally, and all the code can assume invariants are true. For instance, we know that the `i`-th entry in `entries` slice necessarily has `entries[i].index == prev.index + 1 + i`, so we don't have to additionally check for this.

I think the comment already explains this well, but please suggest a better phrasing if it doesn't.

pav-kv:
@ahrtr Addressed actionable comments. Thanks for the review!

ahrtr (approved):
LGTM with a minor comment

Great work. Thanks.

raft.go
```diff
@@ -1745,12 +1745,15 @@ func stepFollower(r *raft, m pb.Message) error {
 }
 
 func (r *raft) handleAppendEntries(m pb.Message) {
-	prev := entryID{term: m.LogTerm, index: m.Index}
-	if prev.index < r.raftLog.committed {
+	// TODO(pav-kv): construct logSlice up the stack next to receiving the
+	// message, and validate it before taking any action (e.g. bumping term).
+	a := logSlice{term: m.Term, prev: entryID{term: m.LogTerm, index: m.Index}, entries: m.Entries}
```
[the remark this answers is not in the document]
pav-kv: Added a `logSliceFromMsgApp` func, because it should specifically be a `MsgApp` message.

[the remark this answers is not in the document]
ahrtr: > because it should specifically be a `MsgApp` message.

In that case, we should add a verification each time when the function is called in case it's misused.. The verification should only be executed in test. Let me add a generic verification framework/utility in a separate PR.

Issue #146, opened 2024-01-31
Revisit tests


pav-kv:
This is an umbrella issue for all kinds of unit-test improvements. Unit-tests in this repository are significantly outdated, and need a revamp. For example:

- [raft_test.go](https://github.com/etcd-io/raft/blob/main/raft_test.go)
- [rawnode_test.go](https://github.com/etcd-io/raft/blob/main/rawnode_test.go)

Examples of things to fix/improve:

- [ ] Fix [incorrect initialization](https://github.com/pav-kv/raft/blob/d14e61bd82582490766aef1d4b13986fa2573a53/raft_test.go#L1569-L1594) of nodes, make a standard way to do this. Some tests change `RawNode / raft` structs arbitrarily, which breaks internal invariants and makes tests prone to false failures on any change.
- [ ] Eliminate boilerplate. For example, all [assertions](https://github.com/pav-kv/raft/blob/d14e61bd82582490766aef1d4b13986fa2573a53/raft_test.go#L1707-L1718) should use `require/assert` packages.
- [ ] Introduce helpers, to eliminate more boilerplate. For example, [entry slices](https://github.com/pav-kv/raft/blob/d14e61bd82582490766aef1d4b13986fa2573a53/raft_test.go#L4515-L4739) don't need to be copy-pasted many times, and can be generated with a simple function.
- [ ] Introduce test-only invariant checks that happen transparently in all tests. An example invariant: `r.Term >= r.raftLog.lastTerm()`. Tests that manipulate `RawNode` structs directly may accidentally break invariants, and end up testing arbitrary behaviour rather than real one.

MrDXY:
Hi @pav-kv, For the code you mentioned in "Eliminate Boilerplate", I am interested in improving it and have the time to do so. If I understand correctly, the following code 
```
if sm.state != tt.wstate {
        t.Errorf("#%d.%d state = %v , want %v", i, j, sm.state, tt.wstate)
}
```
Should it be changed into something like this?
```
assert.Equal(t, tt.wstate, sm.state, "#%d.%d state = %v , want %v", i, j, sm.state, tt.wstate)
```
If you were expecting that, I would like to work on this as a starting point to contribute to etcd.

pav-kv:
@MrDXY Yes, the idea is right. Feel free to send a PR.

In most cases we don't need the formatted text part. It's enough to do this:

```golang
assert.Equal(t, tt.wstate, sm.state)
```

[...]

PR #147, opened 2024-01-31
tests: add helpers for generating entry slices

pav-kv:
Part of #146

### DIFF
--- a/log_test.go
+++ b/log_test.go
@@ -24,26 +24,26 @@ import (
 )
 
 func TestFindConflict(t *testing.T) {
-	previousEnts := []pb.Entry{{Index: 1, Term: 1}, {Index: 2, Term: 2}, {Index: 3, Term: 3}}
+	previousEnts := index(1).terms(1, 2, 3)
 	tests := []struct {
 		ents      []pb.Entry
 		wconflict uint64
 	}{
 		// no conflict, empty ent
-		{[]pb.Entry{}, 0},
+		{nil, 0},
 		// no conflict
-		{[]pb.Entry{{Index: 1, Term: 1}, {Index: 2, Term: 2}, {Index: 3, Term: 3}}, 0},
-		{[]pb.Entry{{Index: 2, Term: 2}, {Index: 3, Term: 3}}, 0},
-		{[]pb.Entry{{Index: 3, Term: 3}}, 0},
+		{index(1).terms(1, 2, 3), 0},
+		{index(2).terms(2, 3), 0},
+		{index(3).terms(3), 0},
 		// no conflict, but has new entries
-		{[]pb.Entry{{Index: 1, Term: 1}, {Index: 2, Term: 2}, {Index: 3, Term: 3}, {Index: 4, Term: 4}, {Index: 5, Term: 4}}, 4},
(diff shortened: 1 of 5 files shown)


pav-kv:
@serathius PTAL. This is just a mechanical clean-up in tests.

log_test.go
```diff
@@ -1044,3 +1010,23 @@ func mustTerm(term uint64, err error) uint64 {
 	}
 	return term
 }
+
+type index uint64
```
serathius, line 1014: Idea of having clearer test case generations seems good, however I'm not sure about the solution you proposed, it doesn't seem very natural. 

Things like `index(0).terms(1,2,3)` that generate pairs `(index, term)` of `(0, 1), (1, 2), (2, 3)`. Also having a function as part of `to` method seems like you really want to force a single function to handle too many different cases.

It's nice that names are short and concise, but downside is that they don't represent what code really does.

`index` seems like more `fromIndex` or `index.from`. `to` seems like it would better without function and depend on terms provided via a `terms`.

For example replace `index(0).to(2, func(index uint64) uint64 { return index })` with `index.from(0).terms(term.from(0).to(2)...)`

pav-kv: Good point re `func` generator in `to` being ugly. Replaced it with a simple `index(0).termsIn(from, to)`.

So we now have 2 generators:

```golang
index(1).terms(1, 2, 2, 3) // the terms are enumerated explicitly
index(1).termsIn(100, 200) // consecutive terms [100, 200)
```

I like the conciseness of `index` better than `fromIndex`, it's kinda the point - for writing less boilerplate in tests. I fixed a slight obscurity of this name by adding some comments to all generators.

serathius: > Good point re func generator in to being ugly. Replaced it with a simple index(0).termsIn(from, to).

Nice, however it would be nicer to follow etcd name convention `index(0).termRange(from, to)`

pav-kv: Done.

raft_paper_test.go
```diff
@@ -832,17 +793,18 @@ func TestVoter(t *testing.T) {
...
 		// candidate higher logterm
-		{[]pb.Entry{{Term: 1, Index: 1}}, 2, 1, false},
-		{[]pb.Entry{{Term: 1, Index: 1}}, 2, 2, false},
+		{index(1).terms(1), 2, 1, false},
+		{index(1).terms(1), 2, 2, false},
 		{[]pb.Entry{{Term: 1, Index: 1}, {Term: 1, Index: 2}}, 2, 1, false},
 		// voter higher logterm
-		{[]pb.Entry{{Term: 2, Index: 1}}, 1, 1, true},
-		{[]pb.Entry{{Term: 2, Index: 1}}, 1, 2, true},
-		{[]pb.Entry{{Term: 2, Index: 1}, {Term: 1, Index: 2}}, 1, 1, true},
```
pav-kv, line 845: This line was incorrect: terms in a log can not regress. Added two correct tests instead.

pav-kv (commented):
@serathius I converted a few more test files.

Once happy, please merge. Thank you.

serathius (approved):
cc @ahrtr

ahrtr (approved):
LGTM

thx for simplifying the test cases.

Issue #148, opened 2024-01-31
Probe fom index less than match


joshuazh-x:
I'm learning probe mechanism and notice there is chance for replication starting from an index less than match index in probe state, unlike that in replicate state. For example, a rejected probe response got delayed and is received when the leader falls into another probe state. This is a rare case of cause.

I'm not sure if this is by design. But to my understanding, allowing such a smaller index implies we want to tolerate loss of already replicated log entries. 

```
func (pr *Progress) MaybeDecrTo(rejected, matchHint uint64) bool {
	if pr.State == StateReplicate {
		// The rejection must be stale if the progress has matched and "rejected"
		// is smaller than "match".
		if rejected <= pr.Match {
			return false
		}
		// Directly decrease next to match + 1.
		//
		// TODO(tbg): why not use matchHint if it's larger?
		pr.Next = pr.Match + 1
		return true
	}

	// The rejection must be stale if "rejected" does not match next - 1. This
	// is because non-replicating followers are probed one entry at a time.
	if pr.Next-1 != rejected {
		return false
	}

	pr.Next = max(min(rejected, matchHint+1), 1)
	pr.MsgAppFlowPaused = false
	return true
}
```

pav-kv:
Are you saying that this line sets `pr.Next` to be less than `pr.Match`?

```golang
pr.Next = max(min(rejected, matchHint+1), 1)
```

joshuazh-x:
> Hi @joshuazh-x. The probe/replicate state distinction are heuristics. Regardless of what they do, it does not impact `raft` safety properties. Basically, probe/replicate try to guess what's the good `MsgApp` message to send. But ultimately, the follower gates these messages into the storage and rejects if they are inconsistent with the log state.
> 
> > But to my understanding, allowing such a smaller index implies we want to tolerate loss of already replicated log entries.
> 
> Could you explain a little what you mean by loss? Do you mean we would be sending entries that are already in `Match` a second time?
I meant we don't need to replicate entries preceding match index, The only necessity to do so is when follower would have lost replicated log entries during the same term, which is impossible.

joshuazh-x:
> Are you saying that this line sets `pr.Next` to be less than `pr.Match`?
> 
> ```go
> pr.Next = max(min(rejected, matchHint+1), 1)
> ```
> 
> If so, that would not be by design, rather a bug. The invariant is that `Next > Match`.
> 
[...]

PR #149, opened 2024-02-01
raft: next index shall be larger than match index.

joshuazh-x:
Leader shall always replicate its log from a index larger than corresponding match index. Note that this is not a correctness issue, but rather an optimization in case of message reordering.

Related: #148

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -4824,6 +4824,58 @@ func votedWithConfig(configFunc func(*Config), vote, term uint64) *raft {
 	return sm
 }
 
+func TestLogReplicationWithReorderedMessage(t *testing.T) {
+	r1 := newTestRaft(1, 10, 1, newTestMemoryStorage(withPeers(1, 2)))
+	r1.becomeCandidate()
+	r1.becomeLeader()
+	r1.readMessages()
+	r1.trk.Progress[2].BecomeReplicate()
+
+	r2 := newTestRaft(2, 10, 1, newTestMemoryStorage(withPeers(1, 2)))
+
+	// r1 sends 2 MsgApp messages to r2.
+	mustAppendEntry(r1, pb.Entry{Data: []byte("somedata")})
+	r1.sendAppend(2)
+	req1 := expectOneMessage(t, r1)
+	mustAppendEntry(r1, pb.Entry{Data: []byte("somedata")})
+	r1.sendAppend(2)
+	req2 := expectOneMessage(t, r1)
+
+	// r2 receives the second MsgApp first due to reordering.
+	r2.Step(req2)
+	resp2 := expectOneMessage(t, r2)
+	// r2 rejects req2
+	require.True(t, resp2.Reject)
+	require.Equal(t, uint64(0), resp2.RejectHint)
+	require.Equal(t, uint64(2), resp2.Index)
+
+	// r2 handles the first MsgApp and responses to r1.
+	// And r1 updates match index accordingly.
+	r2.Step(req1)
+	m := expectOneMessage(t, r2)
+	require.False(t, m.Reject)
+	require.Equal(t, uint64(2), m.Index)
+	r1.Step(m)
+	m = expectOneMessage(t, r1)
+	require.Equal(t, uint64(2), r1.trk.Progress[2].Match)
+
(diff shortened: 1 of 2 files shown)


raft_test.go
```diff
@@ -4824,6 +4824,57 @@ func votedWithConfig(configFunc func(*Config), vote, term uint64) *raft {
...
+	// now r1 receives resp2
+	r1.Step(resp2)
+	m = expectOneMessage(t, r1)
+	// r1 shall re-resend from match index even if resp2's reject hint is less than matching index
+	require.Equal(t, r1.trk.Progress[2].Match, m.Index)
+}
+
+func expectOneMessage(t *testing.T, r *raft) pb.Message {
+	msgs := r.readMessages()
+	require.Len(t, msgs, 1, "expecte one message")
```
pav-kv, line 4874: expected?

tracker/progress.go
```diff
@@ -207,11 +207,13 @@ func (pr *Progress) MaybeDecrTo(rejected, matchHint uint64) bool {
...
 	// The rejection must be stale if "rejected" does not match next - 1. This
 	// is because non-replicating followers are probed one entry at a time.
+	// The check is a best effort assuming message reordering is rare.
 	if pr.Next-1 != rejected {
 		return false
 	}
 
-	pr.Next = max(min(rejected, matchHint+1), 1)
+	// Next index shall always be larger than match index.
+	pr.Next = max(min(rejected, matchHint+1), pr.Match+1)
```
pav-kv, line 216: Did the test fail before this change? I.e. you verified that the bug was real?

pav-kv (approved):
LGTM % a couple minor things

tracker/progress.go
Tony0X5A: Yes, the test will fail at the last step of verification without this change.

joshuazh-x: The test will fail without this change. The issue was actually found in trace validation when I enforce the model to always send MsgApp with larger match index. Then I reproduce this with this test case.

raft_test.go
joshuazh-x: My typo. I've fixed it in new commit.

pav-kv:
@joshuazh-x Thanks for the fix.

@ahrtr Could you review/merge? Thank you.

raft_test.go
```diff
@@ -4824,6 +4824,57 @@ func votedWithConfig(configFunc func(*Config), vote, term uint64) *raft {
...
+func TestLogReplicationWithReorderedMessage(t *testing.T) {
+	r1 := newTestRaft(1, 10, 1, newTestMemoryStorage(withPeers(1, 2)))
+	r1.becomeCandidate()
+	r1.becomeLeader()
+	r1.readMessages()
+	r1.trk.Progress[2].BecomeReplicate()
+
+	r2 := newTestRaft(2, 10, 1, newTestMemoryStorage(withPeers(1, 2)))
+
+	// r1 sends 2 MsgApp messages r2.
```
ahrtr, line 4836:
```suggestion
	// r1 sends 2 MsgApp messages to r2.
```

ahrtr:
LGTM with one nit.

Confirmed that the test case will fail without the fix.

```
--- FAIL: TestLogReplicationWithReorderedMessage (0.00s)
    raft_test.go:4869: 
        	Error Trace:	/Users/wachao/go/src/github.com/joshuazh-x/raft/raft_test.go:4869
        	Error:      	Not equal: 
        	            	expected: 0x2
        	            	actual  : 0x0
        	Test:       	TestLogReplicationWithReorderedMessage
FAIL
exit status 1
FAIL	go.etcd.io/raft/v3	0.565s
```

Probably we should add an invariant check each time when sending a MsgApp message to ensure that the index should always be greater than the Match. It can be addressed in a separate PR.

raft_test.go
joshuazh-x: Fixed the typo.

ahrtr:
Please rebase this PR although github doesn't show any conflict.

joshuazh-x:
> Please rebase this PR although github doesn't show any conflict.

done.

ahrtr (approved):
lgtm

Thanks

PR #152, opened 2024-02-03
raft:improve the lease

1797818494:
When the vote is from the node's leader, it say that the Lease is invalid

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1062,7 +1062,7 @@ func (r *raft) Step(m pb.Message) error {
 	case m.Term > r.Term:
 		if m.Type == pb.MsgVote || m.Type == pb.MsgPreVote {
 			force := bytes.Equal(m.Context, []byte(campaignTransfer))
-			inLease := r.checkQuorum && r.lead != None && r.electionElapsed < r.electionTimeout && r.Lead != m.From
+			inLease := r.checkQuorum && r.lead != None && r.electionElapsed < r.electionTimeout && r.lead != m.From
 			if !force && inLease {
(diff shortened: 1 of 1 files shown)


1797818494 (commented):
it's ok

1797818494:
hi @pavelkalinnikov, could you please review the PR?

raft.go
```diff
@@ -1062,7 +1062,7 @@ func (r *raft) Step(m pb.Message) error {
 	case m.Term > r.Term:
 		if m.Type == pb.MsgVote || m.Type == pb.MsgPreVote {
 			force := bytes.Equal(m.Context, []byte(campaignTransfer))
-			inLease := r.checkQuorum && r.lead != None && r.electionElapsed < r.electionTimeout
+			inLease := r.checkQuorum && r.lead != None && r.electionElapsed < r.electionTimeout && r.lead != m.From
```
ahrtr, line 1065: Did you see a real case that a leader sent vote or pre-vote (`r.lead == m.From`) and when the lease hasn't expired yet (`r.electionElapsed < r.electionTimeout`)?

It will never happen. Please create a case to reproduce it if you still think it's an issue.

ahrtr: FYI. 

https://github.com/etcd-io/raft/blob/2f109970c68e505c223f4d3d5c13a8dcd8f7d528/raft.go#L948-L951

1797818494: Sorry, I mistakenly thought r.electionTimeout is random, but r.randomizedElectionTimeout really is,

PR #151, opened 2024-02-01
log: fixup raftLog initialization

pav-kv:
This PR consolidates the `raftLog` struct initialization into a single return statement. It also makes all unit-tests use the `newLog` constructor function to initialize a log, instead of populating the struct manually.

Changes to the `unstable` / `raftLog` structures should by default not require fixing boilerplate in tests.

Related to #146
Blocks #139

PR #155, opened 2024-02-05
log: clean-up log conflict search

pav-kv:
This PR cleans up `raftLog.maybeAppend` and `raftLog.findConflict` functions, now that they accept the new `logSlice` type as input. This replaces a bunch of ad-hoc arithmetics on the raw `entries` slice with a more understandable "fast-forward" logic on a `logSlice`.

This also removes a redundant `panic` in `maybeAppend`. An equivalent panic is present in the `raftLog.append` method which is triggered under the same conditions. This panic behaviour is tested in `TestLogMaybeAppend`, and hasn't changed.

Related to #144

### DIFF
--- a/log.go
+++ b/log.go
@@ -109,9 +109,6 @@ func (l *raftLog) maybeAppend(a logSlice, committed uint64) (lastnewi uint64, ok
 	if !ok {
 		return 0, false
 	}
-	if match.index < a.lastIndex() && match.index < l.committed {
-		l.logger.Panicf("entry %d is already committed [committed(%d)]", match.index+1, l.committed)
-	}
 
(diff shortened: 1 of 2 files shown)


pav-kv:
@ahrtr This PR is based on top of testing clean-ups in #151 and #147. Only the last 3 commits need to be reviewed as part of this PR. Thank you.

pav-kv:
CC @nvanbenschoten

Specifically, consider the TODOs I've added in `findConflict` method. I wonder if this would give any significant boost to log append performance. Currently, we (re-)scan the log upon every append request. I suppose though the the win wouldn't be large, because the common/happy path is that the append is right at the tip of the log.

PR #161, opened 2024-02-08
rawnode: expose per-follower MsgApp message stream

pav-kv:
Part of #130

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1527,7 +1527,7 @@ func stepLeader(r *raft, m pb.Message) error {
 		}
 	case pb.MsgHeartbeatResp:
 		pr.RecentActive = true
-		pr.MsgAppFlowPaused = false
+		pr.PauseMsgAppProbes(false)
 		r.maybeSendAppend(m.From, pr)
 
 		if r.readOnly.option != ReadOnlySafe || len(m.Context) == 0 {
@@ -1561,7 +1561,7 @@ func stepLeader(r *raft, m pb.Message) error {
 		// If snapshot finish, wait for the MsgAppResp from the remote node before sending
 		// out the next MsgApp.
 		// If snapshot failure, wait for a heartbeat interval before next try
-		pr.MsgAppFlowPaused = true
+		pr.PauseMsgAppProbes(true)
 	case pb.MsgUnreachable:
 		// During optimistic replication, if the remote becomes unreachable,
 		// there is huge probability that a MsgApp is lost.
@@ -1598,7 +1598,7 @@ func stepLeader(r *raft, m pb.Message) error {
 			r.sendTimeoutNow(leadTransferee)
 			r.logger.Infof("%x sends MsgTimeoutNow to %x immediately as %x already has up-to-date log", r.id, leadTransferee, leadTransferee)
 		} else {
-			pr.MsgAppFlowPaused = false
+			pr.PauseMsgAppProbes(false)
 			r.maybeSendAppend(leadTransferee, pr)
 		}
 	}
--- a/raft.go
+++ b/raft.go
@@ -626,7 +626,8 @@ func (r *raft) getMessages(to uint64, fc FlowControl, buffer []pb.Message) []pb.
 	}
 	pr := r.trk.Progress[to]
 	buf := msgBuf(buffer)
-	r.maybeSendAppendBuf(to, pr, &buf)
+	for r.maybeSendAppendBuf(to, pr, &buf) {
+	}
 	return buf
 }
 
@@ -648,6 +649,10 @@ func (r *raft) maybeSendAppend(to uint64, pr *tracker.Progress) bool {
 		return false
 	}
 	return r.maybeSendAppendBuf(to, pr, &r.msgs)
+}
+
+func (r *raft) appendsReady(pr *tracker.Progress) bool {
+	return pr.ShouldSendMsgApp(r.raftLog.lastIndex(), r.raftLog.committed)
 }
 
 // maybeSendAppendBuf implements maybeSendAppend, and puts the messages into the
--- a/raft.go
+++ b/raft.go
@@ -640,7 +640,7 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 		Entries: ents,
 		Commit:  r.raftLog.committed,
 	})
-	pr.UpdateOnEntriesSend(len(ents), uint64(payloadsSize(ents)))
+	pr.SentEntries(len(ents), uint64(payloadsSize(ents)))
 	return true
 }
 
--- a/raft.go
+++ b/raft.go
@@ -599,6 +599,10 @@ func (r *raft) sendAppend(to uint64) {
 // argument controls whether messages with no entries will be sent
 // ("empty" messages are useful to convey updated Commit indexes, but
 // are undesirable when we're sending multiple messages in a batch).
+//
+// TODO(pav-kv): make invocation of maybeSendAppend stateless. The Progress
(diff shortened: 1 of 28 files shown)


raft.go
```diff
@@ -218,6 +218,21 @@ type Config struct {
 	// throughput limit of 10 MB/s for this group. With RTT of 400ms, this drops
 	// to 2.5 MB/s. See Little's law to understand the maths behind.
 	MaxInflightBytes uint64
+	// DisableEagerAppends makes raft hold off constructing log append messages in
+	// response to Step() calls. The messages can be collected via a separate
+	// MessagesTo method.
```
sumeerbhola, line 223: I'm not very clear on usage, specifically compared to how we currently call `RawNode.Ready()` in CockroachDB, and then send the outbound messages in `Ready.Messages`.

If this is set to true, what will `Ready.Messages` contain? Is it only `MsgStorageAppend` and `MsgStorageApply` messages to the local store?

How will the user know to call `MessagesTo` -- presumably the user has to know that there is something available to send to a replica? Is the idea that the user will retrieve `tracker.Progress` and somehow know that there are entries available >= `Progress.Next`, and then decide to call `MessagesTo`? If it gets back `(Progress.Next, j]` and sends them, will it then call `Progress.SentEntries(j)` that will update `Progress.Next` to `j+1`? Isn't `Progress` just a struct, so how will this `Progress` state change go back into the `RawNode`?

Does etcd/raft assume that whatever has been retrieved in `MessagesTo` is the responsibility of the caller to reliably deliver, or does it itself retry? That is, who is responsible for ensuring that `(Progress.Match,Progress.Next)` is delivered? I am also murky on whether everything `<= Progress.Match` has indeed been delivered, because of the following comment in the CockroachDB code that says:
```
		// In state ProgressStateReplicate, the Match index is optimistically
		// updated whenever a message is *sent* (not received). Due to Raft
		// flow control, only a reasonably small amount of data can be en
		// route to a given follower at any point in time.
```

sumeerbhola (commented):
some very basic questions, due to unfamiliarity with etcd/raft (and fading memory of the Raft thesis).

raft.go
pav-kv: `Ready()` currently returns all messages: this includes appends/snapshots to followers, appends / applies to local storage, heartbeats, etc.

With `DisableEagerAppends` it would still return all these messages, except `MsgApp` and snapshots. These would go through the `MessagesTo` interface. In CRDB, we would still need to call and act on `Ready`. But we would also call `MessagesTo` next to calling `Ready` (or may do so separately / on a different schedule, because it's not necessary to pair with `Ready` calls).

We would not use / update `Progress` structs, raft internally does so when it emits messages.

> How will the user know to call MessagesTo -- presumably the user has to know that there is something available to send to a replica?

We also don't necessarily need to sneak peek into `Progress`. The `MessageTo` calls will return appends if they need to happen, and won't return anything if `Progress` indicates that the follower is up-to-date or throttled.

> Does etcd/raft assume that whatever has been retrieved in `MessagesTo` is the responsibility of the caller to reliably deliver, or does it itself retry?

The expectation is that these messages are going to be sent. There is no reliability expectation, raft assumes that all messages can be dropped. Raft will retry things periodically, and if it detects that the appends message stream was broken, it will resend the portion of the log that was not acked.

Entries returned from `MessagesTo` will now be considered in-flight by raft.

> because of the following comment in the CockroachDB code that says

The comment is inaccurate. It is the `Next` index that's optimistically updated when a message is sent. The `Match` index is a guaranteed match (because the follower replied with a `MsgAppResp` acking log indices up to `Match`).

pav-kv: One thing we need to know to control this is the IDs of the followers, because `MessagesTo` takes the node ID of the recepient.

One way to know the IDs is iterating the `Progress` map, or looking at the current config. We need to be careful with config changes here though - I haven't considered the implications here yet.

tracker/progress.go
```diff
@@ -28,7 +28,19 @@ import (
...
 // certain State. All of this isn't ideal.
 type Progress struct {
-	Match, Next uint64
+	// Match is the log index up to which the follower's log matches the leader's.
+	Match uint64
+	// Next is the log index of the next entry to send to this follower. All
+	// entries with indices in (Match,Next) interval are already in flight.
+	//
+	// Invariant: 0 <= Match < Next.
+	Next uint64
```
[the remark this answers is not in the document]
pav-kv: Yeah, sounds right. Need to double-check the invariants here hold at at all times.

In CRDB, these will be promptly promoted to (10,11) after the snapshot initialization.

raft.go
pav-kv: Yes, the retries happen like you described, transparently to the application-level caller. The application just needs to send the messages.

Best results are achieved if these messages are delivered in order though (that's also the case today), because each `MsgApp` builds on top of the previous ones. But if any message is dropped and the flow is "interrupted", raft will learn this, rollback the `Next` and retry.

nvanbenschoten: > How will the user know to call MessagesTo

I'm also curious about this. Will this be hooked in to `RawNode.HasReady`? Will we introduce a new `HasMessagesTo(FlowControl)` API? What will indicate to CockroachDB that a replica needs to be scheduled on the raft scheduler?

raft.go
```diff
@@ -372,6 +393,10 @@ type raft struct {
 	// Messages in this list have the type MsgAppResp, MsgVoteResp, or
 	// MsgPreVoteResp. See the comment in raft.send for details.
 	msgsAfterAppend []pb.Message
+	// disableEagerAppends instructs append message construction and sending until
+	// the Ready() call. This improves batching and allows better resource
```
nvanbenschoten, line 397: > until the Ready() call

Until the MessagesTo() call?

rawnode.go
```diff
@@ -136,6 +136,29 @@ func (rn *RawNode) Ready() Ready {
 	return rd
 }
 
+// FlowControl tunes the volume and types of messages that GetMessages call can
+// return to the application.
+type FlowControl struct {
```
nvanbenschoten, line 141: Did you explore pushing `FlowControl` as an interface into `etcd/raft`, similar to library's relationship with the `Storage` interface? I would imagine that this would allow us to hide the interactions with the optional flow control without needing to change the external interface of the library.

We probably still would want a `DisableEagerAppends` option so that the library could be configured to only consult `FlowControl` on calls to `Ready()`/`HasReady()`. These functions would then loop over the peers and construct messages for each according to the `FlowControl` state. Pushing this into `etcd/raft` would also facilitate fast-paths for replica states (e.g. followers) that cannot send MsgApps and need not consult `FlowControl`.

pav-kv: Both ways do extend the surface of the library, but interfacing `FlowControl` makes it indirect and implicit. I'm slightly in favour of avoiding this inversion of control, although I'm still exploring options.

pav-kv:
Superseded by https://github.com/cockroachdb/cockroach/pull/125002 in CRDB.

raft.go
pav-kv: For posterity: the supposed flow here is that the upper layer (e.g. CRDB) tracks all follower flows in `StateReplicate` which are "ready" to send entries. The condition for this is `Next <= raftLog.lastIndex() && !Inflights.Full()`. The tracking is "perfect", in that raft will expose all events that help tracking this info and keeping it consistent with its own `Progress` (e.g. going in/out of `StateLeader` and `StateReplicate`, config changes that remove a follower, etc).

Then, on `Ready` processing, the caller goes through all "ready" flows and calls this method to send messages. The "perfect" tracking helps avoiding wasted calls.

PR #165, opened 2024-02-14
tracker: clean-up Match, Next and MaybeUpdate

pav-kv:
This PR documents the meaning of `Match` and `Next` field in the `Progress` struct. It also ensures that the invariant is maintained in all code places that initialize or modify these fields.

It then simplifies the `Progress.MaybeUpdate` method, given that the invariant is true before it is invoked.


pav-kv:
@ahrtr PTAL

confchange/confchange.go
```diff
@@ -259,8 +259,8 @@ func (c Changer) initProgress(cfg *tracker.Config, trk tracker.ProgressMap, id u
 		// at all (and will thus likely need a snapshot), though the app may
 		// have applied a snapshot out of band before adding the replica (thus
 		// making the first index the better choice).
-		Next:      c.LastIndex,
 		Match:     0,
+		Next:      max(c.LastIndex, 1), // invariant: Match < Next
```
ahrtr, line 263: Why the `Next` is set to `lastIndex + 1` in reset? Should we make them consistent? 

https://github.com/etcd-io/raft/blob/83d8deca057fef78cf3a86ea70457e4d3255e4da/raft.go#L784

pav-kv: I tried `c.LastIndex + 1` here for consistency, but it breaks a few tests. I think the reason is: setting to `lastIndex + 1` precludes sending a `MsgApp` message instantly (the `maybeSendAppend` treats it as "everything is in flight", and won't send any appends until there is a heartbeat reply). So, here we set it to `LastIndex` to guarantee at least one message send. Also, typically this last entry contains the very config change that we're performing here.

ahrtr: > setting to lastIndex + 1 precludes sending a MsgApp message instantly

Makes sense.

tracker/progress.go
```diff
@@ -115,17 +123,17 @@ func (pr *Progress) BecomeProbe() {
 	if pr.State == StateSnapshot {
 		pendingSnapshot := pr.PendingSnapshot
 		pr.ResetState(StateProbe)
-		pr.Next = max(pr.Match+1, pendingSnapshot+1)
+		pr.Next = max(pr.Match+1, pendingSnapshot+1) // invariant: Match < Next
```
ahrtr, line 126: I see there are comments "`// invariant: Match < Next`" in multiple places. I think we only need to keep the comment in `type Progress struct`, and also add verification (I was planning to do it for a while) to safeguard it instead of adding comment everywhere.

ahrtr (approved):
lgtm with a minor comment

cc @erikgrinaker

tracker/progress.go
pav-kv: Removed the comment from all the places, except the non-trivial `max(c.LastIndex, 1)` in one place (where this invariant was not maintained before this PR).

erikgrinaker (approved):
LGTM, sorry for the holdup.

PR #167, opened 2024-02-20
Do not reset the electionElapsed if the node doesn't grant vote or preVote

ahrtr:
The first commit refactors the existing code without changing any logic. The second commit changes the logic as the PR title mentions. Still need to add a test.

```
// If the local node receives a message with higher term,
// but it doesn't grant the vote (or pre-vote); it turns
// into a follower, but it shouldn't reset the electionElapsed,
// to ensure it has higher priority to start a campaign
// in the next round of election. If we reject a node, it's
// highly likely we will reject it again if it immediately
// campaigns again.
```


ahrtr:
cc @pav-kv @erikgrinaker PTAL

ahrtr:
> I assume the motivation of this change comes from [etcd-io/etcd#17455](https://github.com/etcd-io/etcd/issues/17455).

Yes.



> When such a node does not reset its election timeout, it may quickly start a new campaign and makes the winning candidate step down to follower again before it becomes leader.

This PR doesn't change the local node's frequency to campaign. The PR just prevents any node being affected by other unqualified candidate. 
- This PR only fixes the case when pre-vote isn't enabled. So as long as pre-vote is enabled, there is no any side effect.
- The side effect you mentioned only exist when two nodes campaign at the same time. It's resolved by the randomised election timeout.

ahrtr:
Prefer to the alternative PR https://github.com/etcd-io/raft/pull/169,
- #169 has better readability than this PR
- In this PR, the local node's term has already increased by 1 even it rejects the Vote. So Conceptually it should reset the `electionElapsed`.

pav-kv:
@ahrtr

> In this PR, the local node's term has already increased by 1 even it rejects the Vote. So Conceptually it should reset the electionElapsed.

Why does it conceptually need to reset the timeout? The timeout behaviours are heuristics, so there is no right or wrong ways. We should consider behaviours and do some calculations to compare heuristics.

Do we need to reset `electionTimeout` at all in situations when a non-leader state remains non-leader? E.g. when we `BecomeFollower`, and we were already in `StateFollower`, just with a lower term.

ahrtr:
> Upd: I guess we do need to sometimes reset it, because of [#167 (comment)](https://github.com/etcd-io/raft/pull/167#issuecomment-1955813036).

YES, it's one of the reasons why I prefer to https://github.com/etcd-io/raft/pull/169. The other reason is #169 has better readability. 



> Why does it conceptually need to reset the timeout? The timeout behaviours are heuristics, so there is no right or wrong ways. We should consider behaviours and do some calculations to compare heuristics.
> 
> Do we need to reset `electionTimeout` at all in situations when a non-leader state remains non-leader?

All election time parameters (e.g. `randomizedElectionTimeout`) are supposed to be bind with each term. If the term changes (increases), then we should reset all such time parameters. At least, this is current implementation. If we don't reset the value, and carry over to next term, we don't know what other side effect it may cause apart from https://github.com/etcd-io/raft/pull/167#issuecomment-1955813036. I agree with "`there is no right or wrong ways`", but I tend not to take risk for now.

PR #169, opened 2024-02-22
Delay the next campaign if the node lost the vote

ahrtr:
Delay the next campaign if the node lost the vote. It's highly likely it will also lose next campaign, so it makes more sense to prioritize campaigns by other nodes within the current term.

Alternative to resolve https://github.com/etcd-io/etcd/issues/17455. It has better readability than https://github.com/etcd-io/raft/pull/167.


ahrtr:
> I think a better approach would be to verify if the rejection is due to outdated logs before delaying the next request.

Yes, I was thinking the same. But unfortunately, we don't have such info in the `MsgVoteResp`.



> This approach may potentially increase the downtime of the cluster.

It should be fine. It just delays current node's next campaign. It doesn't change other nodes' campaign timings. Also the `randomizedElectionTimeout` will be reset to the normal value in next term.

ahrtr:
@Cjen1 I think this PR can also resolve what your PR https://github.com/etcd-io/raft/pull/86 is trying to fix, and with minimum change. Please take a look and let me know if you have any comment. thx.

@erikgrinaker @pav-kv PTAL

erikgrinaker:
@ahrtr Hey, sorry for the review lag here, it's been a crazy week. I'll be on PTO for the next week, but maybe @pav-kv can take a look.

raft.go
```diff
@@ -1674,6 +1674,14 @@ func stepCandidate(r *raft, m pb.Message) error {
 			// pb.MsgPreVoteResp contains future term of pre-candidate
 			// m.Term > r.Term; reuse r.Term
 			r.becomeFollower(r.Term, None)
+			// Delay the next campaign if the node lost the vote. It's
+			// highly likely it will also lose next campaign, so it makes
+			// more sense to prioritize campaigns by other nodes within
+			// the current term. Note the `randomizedElectionTimeout` will
+			// be automatically reset in next term.
+			if myVoteRespType == pb.MsgVoteResp {
+				r.randomizedElectionTimeout += r.electionTimeout
```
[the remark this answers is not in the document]
pav-kv: The idea is to skip the next `electionTimeout` to give other nodes a change to win in the term+1 rather than current term?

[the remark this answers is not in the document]
pav-kv: Enabling `PreVote` seems like a better fix for this?

pav-kv:
@ahrtr Are you able to reproduce the problem in https://github.com/etcd-io/etcd/issues/17455 in a datadriven test? What kind of race are we trying to fix?

pav-kv:
This change seems to be concerned with the situation when a candidate loses because of having an outdated log. However, the fix may affect the situation of a split vote. When two candidates duel and both lose campaign because of a split vote, both will bump the election timeout, and the election may stall for a bit.

ahrtr:
> Enabling `PreVote` would address the outdated log problem more directly.

Yes. This PR is fixing the use cases where PreVote isn't enabled at all.

> The idea is to skip the next electionTimeout to give other nodes a chance to win in the term+1 rather than current term?

The idea is to increase the current node's `randomizedElectionTimeout` when it lost the vote. Normally it's in range `[electionTimeout, 2*electionTimeout)`. If a node lost the vote, we intentionally set the range to `[2*electionTimeout, 3*electionTimeout)`, so that other nodes will always campaign before the node, which already lost the vote, within current term.

Let's work with an example. Assuming a cluster has 3 nodes, and one of the node (say node **A**) is lag behind other nodes. When they start to campaign, the node A has a possibility of 33% to start the campaign before other nodes. Its vote will definitely be rejected by other nodes. Then all the 3 nodes will conduct the next round of election fairly. The node A still has a possibility of 33% to start the campaign before other nodes. Eventually it may lead to a long time for a real leader to get elected. It's exactly this PR tries to fix.

Note that `randomizedElectionTimeout` will automatically be reset in next term.



> When two candidates duel and both lose campaign because of a split vote, both will bump the election timeout, and the election may stall for a bit.

Yes, it's true. But I imagine that the possibility of running into such split vote should be far less than 33% mentioned above. Usually we recommend odd number of members in a cluster. But it isn't verified yet. Ideally we should a tool to measure the average recovery time, something like what https://github.com/etcd-io/raft/pull/86 does.

> Are you able to reproduce the problem in https://github.com/etcd-io/etcd/issues/17455 in a datadriven test (the tests in testdata, for TestInteraction)?

The datadriven test might not be a good choice to reproduce & verify the fix. The PR doesn't change the logic, so it hasn't any impact on correctness.

ahrtr:
> When two candidates duel and both lose campaign because of a split vote, both will bump the election timeout, and the election may stall for a bit.

Even in split vote, we do not set all nodes' `randomizedElectionTimeout` to `[2*electionTimeout, 3*electionTimeout)`. Instead we only update the `randomizedElectionTimeout` for the nodes which have lost the vote.

pav-kv:
> The datadriven test might not be a good choice to reproduce & verify the fix. The PR doesn't change the logic, so it hasn't any impact on correctness.

The datadriven test is a good way, IMO. It clearly shows the behaviour. Yes, we're not changing correctness, but we're changing the behaviour. The datadriven harness gives ways to emulate lagging and duelling campaigns, you can cause specific message delivery ordering that repro the problem. There is also a [command](https://github.com/etcd-io/raft/blob/d475d7e4aa8ce5c2b2a42765e9bdeb9689966663/rafttest/interaction_env_handler.go#L100-L107) for setting specific election timeouts which might be useful here.

To demonstrate the effect of this PR, it would be good to 1) in the first commit add a test with the election rejection scenario you're targetting, 2) fix it in the second commit - the diff in the `testdata` file will demonstrate the effect.

> The idea is to increase the current node's randomizedElectionTimeout when it lost the vote. Normally it's in range [electionTimeout, 2*electionTimeout). If a node lost the vote, we intentionally set the range to [2*electionTimeout, 3*electionTimeout), so that other nodes will always campaign before the node, which already lost the vote, within current term.

I'm still not getting the problem. Say, all the node campaign in the `[electionTimeout, 2*electionTimeout)` interval: at t1, t2, t3. Yes, the first node A can lose the election at t1. If the reason for rejected votes was a short log, all the nodes will vote in this same term for another, longer log, at t2 or t3. Node A already moves its election time by at least `electionTimeout`, so next time it will campaign t4 is > t2 and > t3.

ahrtr:
> If the reason for rejected votes was a short log, all the nodes will vote in this same term for another, longer log, at t2 or t3. Upon losing the election, node A already moves its election time by at least `electionTimeout`, so next time it will campaign t4 is > t2 and > t3.

Note this PR is fixing the cases where preVote isn't enabled at all. Each time when any node campaign, it will increase its term by 1. Other nodes will be converted to a follower when receiving the MsgVote. It means even a node lost its vote, all nodes' `randomizedElectionTimeout` and `electionElapsed` will be reset.

https://github.com/etcd-io/raft/blob/d475d7e4aa8ce5c2b2a42765e9bdeb9689966663/raft.go#L1094

https://github.com/etcd-io/raft/blob/d475d7e4aa8ce5c2b2a42765e9bdeb9689966663/raft.go#L774-L777

Cjen1:
@ahrtr Just checking my intuition here. Given there is a lagging node that calls and fails to be elected, you want to bias the retried election to the other replicas?

I think some issues arise when there is asynchrony or the election timeout is roughly the same order of magnitude as the round-trip-time. Specifically this means that before receiving the next `msgVote` the nodes have already sent their own (This causes duelling leaders), and hence cause each retry to be `(2->3) * electionTimeout` rather than `(1->2) * electionTimeout).

For context I've been running some tests in cloudlab in a 3 node cluster with ~50ms latency between nodes.
With 500ms electionTimeout, I'm seeing 2-5 retries before a leader is elected.
(I'm currently calibrating the reckon test setup, so I am not 100% satisified with the reliability of these numbers, or that they weren't statistical flukes)

pav-kv:
@ahrtr I see. You're saying that the nodes who rejected the vote, will become followers and reset the timeout. So my assumption that t4 > t2 and > t3, is incorrect? Then it's possible that the short log A campaigns first again.

I wonder though if, alternatively to the fix in this PR, we should just not reset the election timeout in the nodes who reject a vote and becomeFollower(leader=None). Then they will campaign earlier than the rejected node A tries again.

ahrtr:
> Just checking my intuition here. Given there is a lagging node that calls and fails to be elected, you want to bias the retried election to the other replicas?


Yes. Normally the randomized election timeout is in range `[electionTimeout, 2*electionTimeout)`. If a node lost the vote, we intentionally set the range to `[2*electionTimeout, 3*electionTimeout)`.

But it's **only within one term**. All the time parameters will be reset in next term.

ahrtr:
Added a data-driven test case.   @pav-kv PTAL, suggest to review it commit by commit. @Cjen1 Please also take a look if you have time. Thanks both.

Also responding to your previous comment on split vote, actually the vote result in split vote is `VotePending`. This PR only updates the randomized election timeout when the vote result is `VoteLost`.



> Is it possible to add whether it was rejected because of the log?

No such plan in this PR for now. Also note that one peer's rejection doesn't mean current node will have to necessarily lose the campaign.

ahrtr:
> Also responding to your previous comment on split vote, actually the vote result in split vote is `VotePending`. This PR only updates the randomized election timeout when the vote result is `VoteLost`.

This isn't correct. For split vote, the vote result is `VoteLost` instead of `VotePending`. But my previous comment https://github.com/etcd-io/raft/pull/169#issuecomment-1964584497 still stands.

testdata/vote.txt
```diff
@@ -0,0 +1,97 @@
+# Tests that a node that is behind on the log will lose the vote.
```
pav-kv: I think it would be better to write this test in the first commit in such a way that it demonstrates the problem. As far as I understand, the problem is that a node that failed to be elected, can try campaigning again and lose again. Can we recreate this behaviour before fixing it?

It might be necessary to place this commit after the newly added `datadriven` commands/flags, which would help reproing this issue.

ahrtr: > As far as I understand, the problem is that a node that failed to be elected, can try campaigning again and lose again. Can we recreate this behaviour before fixing it?


We can accurately control the behaviour using the datadriven test. No matter how many times the node 2 retries the campaign, it will always lose the campaign as long as it's lag behind. Actually I thought about letting node 2 retry campaign after it lost the vote, but it doesn't make much sense. There isn't difference between trying 1 time and try N (>=2) times. It also complicates the test case, at least much longer.

As I mentioned previously,  Ideally we should have a tool to measure the average recovery time, something like what #86 does. Probably we should discuss/think about it separately.


> It might be necessary to place this commit after the newly added `datadriven` commands/flags, which would help reproing this issue.

The new flags & command is only for updating the test case (for the 5th commit). Actually we can even add the new flags & command in a separate PR if you insist, but it seems not a big deal.

testdata/vote.txt
```diff
@@ -0,0 +1,97 @@
...
+  INFO 2 became follower at term 2
+> 2 handling Ready
+  Ready MustSync=false:
+  Lead:0 State:StateFollower
+
+raft-state
+----
+1: StateFollower (Voter) Term:2 Lead:0
+2: StateFollower (Voter) Term:2 Lead:0
+
```
pav-kv, line 90: To demonstrate the difference in before/after behaviour, I propose to add the following here:

1. `set-randomized-election-timeout = electionTimeout` on node 1. This makes the election timeout on node 1 higher than the random timeout no node 2.
2. Tick nodes 1 and a few times. Node 1 will campaign first, and lose / disrupt the cluster again.

After the fix, the behaviour will change: step (2) will cause the node 2 campaign first and win.

testdata/vote.txt
pav-kv: I think we could still make this test in such a way that before/after behaviour is more obvious. See the comment below.

testdata/vote.txt
pav-kv: There might be +-1 error in this idea, but the general idea should be clear.

ahrtr: Roughly understood what you mean, but it seems still not working. Actually there isn't a proper value for `set-randomized-election-timeout`.

If you set it as `2*electionTimeout` (assuming it's 2*5=10) on node 1 here,
- before the fix, then 9 ticks won't trigger the campaign for node 1, but it will trigger campaign for node 2 for sure.
- after the fix, node 2's randomized election timeout is set to [10, 15). 9 ticks won't trigger campaign for either node; 10 ticks will trigger campaign for node 1 for sure, but it may also trigger campaign for node 2 (as its minimum possible randomized election timeout is 10).

What I am trying to demonstrate in the test is: once node 2 lost its vote, it will be de-prioritized on the next campaign, so node 1 will start the vote in the following campaign before node 2.

pav-kv: Try `2*electionTimeout-1` then. It's ok if the timeout ends up the same on both nodes. Just make sure that node 2 campaigns first anyway. You can arrange this by invoking `ready` on node 2 first.

pav-kv: Even more, you don't have to `set-randomized-election-timeout` in the first place. You can do this:

1. tick both nodes `2*electionTimeout-1` times.
2. `process-ready 2`
3. `deliver-msgs 1`
4. `stabilize`

With this sequence, both nodes will campaign, but node 2 tries it first, regardless of the ticking and random timeouts. After all, it doesn't matter which node initiates campaign first - what matters is whose votes are delivered first.

After the fix, this sequence will flip to node 1 campaigning and winning.

ahrtr: > * tick both nodes `2*electionTimeout-1` times.
> * `process-ready 2`
> * `deliver-msgs 1`
> * `stabilize`

Yes, it works & thanks. Updated. I agree it makes the review easier by comparing the test changes.

Cjen1:
> @erikgrinaker @pav-kv PTAL when you have time.
> 
> @Cjen1 do you have any immediate concern?
> 
> Thanks.

I am still concerned about this impacting the minimal viable recovery time, since if leaders duel it will double the length of each iteration (afaict).

I fully understand that this isn't a massive concern since the recommendation is to minimize any erroneous elections.

Is there an easy way for me to build a binary to test it in reckon? Otherwise if you swap out the etcd binary in reckon that will also work.

ahrtr:
> I'm concerned with a duelling leaders situation.



> 3\. So now there is a temporary election stall? If we tick both nodes `electionTimeout` times, nobody will campaign.

It's true if there are only two nodes in the cluster and the two nodes are campaigning at exactly the same time.
- Firstly, it's low possibility due to the randomized election timeout. But as I mentioned in https://github.com/etcd-io/raft/pull/169#issuecomment-1964573546, a failed node still has a possibility of 33% to start the next campaign before other nodes without this PR.
- Secondly, usually we recommend odd number of cluster members, e.g 3, 5, 7 etc. As mentioned in https://github.com/etcd-io/raft/pull/169#issuecomment-1964584497, we don't delay all nodes'(or members') election timeout, we only delay the nodes which have failed the campaign. In other nodes, usually there are always some nodes, of which the election timeout isn't updated/delayed.




> Is there an easy way for me to build a binary to test it in reckon? Otherwise if you swap out the etcd binary in reckon that will also work.

Is it enough to provide an etcd binary with this PR? I am happy to help on the test.

Cjen1:
I've run this PR through [reckon](https://github.com/cjen1/reckon), using 100 leader loss events for each config, varying the number of nodes (3 and 5) and the election timeout.

fault-aggregate

I think the result is that this PR has no significant effect on the average time to recovery of etcd.
There are some subtle changes in the tail for example I think there is evidence that it reduces the number of repeated elections, but also prolongs each of these rounds, but I think it would require some analytical techniques to properly pull that apart.

erikgrinaker:
I share @pav-kv's concerns in https://github.com/etcd-io/raft/pull/169#issuecomment-1983994248, and also the concerns voiced in #167. Spurious election disruption is a known weakness when not using pre-vote and checkquorum, and it sounds like enabling those would adequately address https://github.com/etcd-io/etcd/issues/17455. I think we should recommend employing those as a first measure, and only tweak election timeout policies when we see significant problems with those enabled.

ahrtr:
> I think the result is that this PR has no significant effect on the average time to recovery of etcd.
> There are some subtle changes in the tail for example I think there is evidence that it reduces the number of repeated elections, but also prolongs each of these rounds, but I think it would require some analytical techniques to properly pull that apart.

Thanks for the test result. My dev branch is based on etcd release-3.5, while your baseline is based on release-3.4. It might be another reason for the subtle change.



> Spurious election disruption is a known weakness when not using pre-vote and checkquorum, and it sounds like enabling those would adequately address [etcd-io/etcd#17455](https://github.com/etcd-io/etcd/issues/17455).

The PR is trying to mitigate such kind of known weakness. Not all applications enable pre-vote. Note etcd release-3.4 is a stable release, it has never enabled pre-vote. 

As mentioned above https://github.com/etcd-io/raft/pull/169#issuecomment-1964573546, a lag node always has **33%** possibility (in a 3-node cluster) to start the campaign before other nodes without this PR.  The possibility is **50%** if there are only two available nodes.

But the possibility of split vote in a 3 node cluster is below based on raft paper 9.2:
- about 3% when l = 10% (l: network latency / election timeout is 10%);
- about 10% when  l = 20%
- about 18% when l = 10% and with 2 available nodes;
- about 35% when l = 20% and with 2 available nodes;

Obviously 33% vs 3-10%, or 50% vs 18-35%, this PR improves the situation.

PR #171, opened 2024-02-26
tracker: track in-flight commit index

pav-kv:
This commit adds a `Progress.sentCommit` field tracking the highest commit index which the leader sent to the follower. It is used to distinguish cases when a commit index update needs or doesn't need to be sent to a follower.

Touches #131

### DIFF
--- a/tracker/progress.go
+++ b/tracker/progress.go
@@ -40,6 +40,14 @@ type Progress struct {
 	// In StateSnapshot, Next == PendingSnapshot + 1.
 	Next uint64
 
+	// sentCommit is the highest commit index in flight to the follower.
+	//
+	// Generally, it is monotonic, but con regress in some cases, e.g. when
+	// converting to `StateProbe` or when receiving a rejection from a follower.
+	//
+	// In StateSnapshot, sentCommit == PendingSnapshot == Next-1.
+	sentCommit uint64
+
 	// State defines how the leader should interact with the follower.
 	//
 	// When in StateProbe, leader sends at most one replication message
@@ -131,6 +139,7 @@ func (pr *Progress) BecomeProbe() {
 		pr.ResetState(StateProbe)
 		pr.Next = pr.Match + 1
 	}
+	pr.sentCommit = min(pr.sentCommit, pr.Next-1)
 }
 
 // BecomeReplicate transitions into StateReplicate, resetting Next to Match+1.
@@ -145,6 +154,7 @@ func (pr *Progress) BecomeSnapshot(snapshoti uint64) {
 	pr.ResetState(StateSnapshot)
 	pr.PendingSnapshot = snapshoti
 	pr.Next = snapshoti + 1
+	pr.sentCommit = snapshoti
 }
 
 // UpdateOnEntriesSend updates the progress on the given number of consecutive
@@ -172,6 +182,21 @@ func (pr *Progress) UpdateOnEntriesSend(entries int, bytes uint64) {
 	default:
 		panic(fmt.Sprintf("sending append in unhandled state %s", pr.State))
 	}
+}
+
+// CanBumpCommit returns true if sending the given commit index can potentially
+// advance the follower's commit index.
+func (pr *Progress) CanBumpCommit(index uint64) bool {
+	// Sending the given commit index may bump the follower's commit index up to
+	// Next-1 in normal operation, or higher in some rare cases. Allow sending a
+	// commit index eagerly only if we haven't already sent one that bumps the
+	// follower's commit all the way to Next-1.
+	return index > pr.sentCommit && pr.sentCommit < pr.Next-1
+}
+
+// SentCommit updates the sentCommit.
+func (pr *Progress) SentCommit(commit uint64) {
+	pr.sentCommit = commit
 }
 
 // MaybeUpdate is called when an MsgAppResp arrives from the follower, with the
@@ -209,6 +234,8 @@ func (pr *Progress) MaybeDecrTo(rejected, matchHint uint64) bool {
 		//
 		// TODO(tbg): why not use matchHint if it's larger?
 		pr.Next = pr.Match + 1
+		// Regress the sentCommit since it unlikely has been applied.
+		pr.sentCommit = min(pr.sentCommit, pr.Next-1)
 		return true
 	}
 
@@ -220,6 +247,8 @@ func (pr *Progress) MaybeDecrTo(rejected, matchHint uint64) bool {
 	}
 
 	pr.Next = max(min(rejected, matchHint+1), pr.Match+1)
+	// Regress the sentCommit since it unlikely has been applied.
+	pr.sentCommit = min(pr.sentCommit, pr.Next-1)
 	pr.MsgAppFlowPaused = false
 	return true
 }
--- a/tracker/progress.go
+++ b/tracker/progress.go
@@ -35,6 +35,9 @@ type Progress struct {
 	// entries with indices in (Match, Next) interval are already in flight.
 	//
 	// Invariant: 0 <= Match < Next.
+	// NB: it follows that Next >= 1.
+	//
+	// In StateSnapshot, Next == PendingSnapshot + 1.
 	Next uint64
 
 	// State defines how the leader should interact with the follower.
@@ -141,6 +144,7 @@ func (pr *Progress) BecomeReplicate() {
 func (pr *Progress) BecomeSnapshot(snapshoti uint64) {
 	pr.ResetState(StateSnapshot)
 	pr.PendingSnapshot = snapshoti
+	pr.Next = snapshoti + 1
 }
 
 // UpdateOnEntriesSend updates the progress on the given number of consecutive
--- a/tracker/progress.go
+++ b/tracker/progress.go
@@ -157,12 +157,12 @@ func (pr *Progress) BecomeSnapshot(snapshoti uint64) {
 	pr.sentCommit = snapshoti
 }
 
-// UpdateOnEntriesSend updates the progress on the given number of consecutive
-// entries being sent in a MsgApp, with the given total bytes size, appended at
-// log indices >= pr.Next.
+// SentEntries updates the progress on the given number of consecutive entries
+// being sent in a MsgApp, with the given total bytes size, appended at log
+// indices >= pr.Next.
 //
 // Must be used with StateProbe or StateReplicate.
-func (pr *Progress) UpdateOnEntriesSend(entries int, bytes uint64) {
+func (pr *Progress) SentEntries(entries int, bytes uint64) {
 	switch pr.State {
 	case StateReplicate:
 		if entries > 0 {
--- a/raft.go
+++ b/raft.go
@@ -644,7 +644,7 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 		Entries: ents,
 		Commit:  r.raftLog.committed,
 	})
-	pr.UpdateOnEntriesSend(len(ents), uint64(payloadsSize(ents)))
+	pr.SentEntries(len(ents), uint64(payloadsSize(ents)))
 	pr.SentCommit(r.raftLog.committed)
 	return true
 }
--- a/raft.go
+++ b/raft.go
@@ -599,6 +599,10 @@ func (r *raft) sendAppend(to uint64) {
 // argument controls whether messages with no entries will be sent
 // ("empty" messages are useful to convey updated Commit indexes, but
 // are undesirable when we're sending multiple messages in a batch).
+//
+// TODO(pav-kv): make invocation of maybeSendAppend stateless. The Progress
+// struct contains all the state necessary for deciding whether to send a
+// message.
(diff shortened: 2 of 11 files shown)


pav-kv:
@ahrtr PTAL. Forked this from #134. This reduces the amount of unnecessary empty `MsgApp` messages, see the effect on the tests in `testdata`.

pav-kv:
@serathius Can I ask you to review this too? CRDB reviewers are OOO this week.

pav-kv:
@bdarnell This addresses your [comment](https://github.com/etcd-io/raft/pull/134#issuecomment-1964452359) to a good degree, without changing the protocol.

tracker/progress.go
```diff
@@ -127,13 +139,15 @@ func (pr *Progress) BecomeProbe() {
 	} else {
 		pr.ResetState(StateProbe)
 		pr.Next = pr.Match + 1
+		pr.pendingCommit = min(pr.pendingCommit, pr.Next-1)
```
ahrtr, line 142: Why does not execute this for the if case `if pr.State == StateSnapshot`?

pav-kv: Bug. Moved this line to the right place.

pav-kv: Actually, `pendingCommit` is not well defined in `StateSnapshot`. We should probably set it to `PendingSnapshot`.

pav-kv: Done. This further reduces the amount of unnecessary messages.

ahrtr (approved):
lgtm

pav-kv:
@serathius Could you take a look please?

ahrtr:
> @ahrtr Are you expecting a second review?

Yes, it's definitely better to have a second review for any behaviour or logic change. We need to minimize any risk as much as possible.

For any mechanical change,  such as just renaming or simple code refactor, it's OK to merge with only one approval.  Such as https://github.com/etcd-io/raft/pull/172

> could you start reviewing #134

It's in my to-do list. Will take a look later.

tracker/progress.go
```diff
@@ -35,8 +35,22 @@ type Progress struct {
...
 	// Invariant: 0 <= Match < Next.
+	// NB: it follows that Next >= 1.
 	Next uint64
 
+	// pendingCommit is the highest commit index <= Next-1 in flight to the
+	// follower.
+	//
+	// The actual in-flight commit index can be higher, but we track only up to
+	// Next-1, because higher indices can be ignored by the peer or reduced to
+	// Next-1 when all the in-flight entries are appended. When Next regresses,
```
bdarnell, line 46: This is an incomplete leader-side copy of the follower logic that caps the commit index. We may pessimistically lower pendingCommit based on an update to Next, which would result in an additional message to convey an updated commit index. This PR tests only the reverse case, when the tracking of pendingCommit *prevents* the sending of a message. 

In general this feels fragile. The goal of my other comment was to get away from heuristics and logic that is duplicated between leader and follower side. I wanted some way for the follower to say "I believe the commit index is X" instead of the leader saying that "when Next updates, set Progress.pendingCommit".

pav-kv: I agree, this logic would be more complete if the leader tracked the follower's commit index based on what the follower says.

This is similar to the `Match/Next` tracking for the log state:

1. `Match` tracks the log index that is stored durably on the follower (confirmed by a follower->leader message).
2. `Next` tracks the optimistic log index - all entries in `(Match, Next)` are in-flight to the follower.

For commit index, we should similarly have:

1. `Committed` index that is stored durably on the follower (confirmed by a follower->leader message).
2. `pendingCommit >= Commit` index that is in-flight to the follower.

This PR implements part (2), and improves status quo. Implementing part (1) requires a protocol change. This is captured by #131.

For the purposes of flow control improvements that I'm making, adding only part (2) for now is good enough.

pav-kv: > We may pessimistically lower pendingCommit based on an update to Next, which would result in an additional message to convey an updated commit index.

Note that we send the actual commit index, not the capped `pendingCommit`. We cap `pendingCommit` only here in the tracker.

pav-kv:
@serathius ping

pav-kv:
@ptabor @spzala Could you review/approve please?

tracker/progress.go
```diff
@@ -35,8 +35,22 @@ type Progress struct {
...
+	// pendingCommit regresses too.
+	//
+	// In StateSnapshot, pendingCommit == PendingSnapshot == Next-1.
+	//
+	// Invariant: 0 <= pendingCommit < Next.
+	pendingCommit uint64
```
erikgrinaker, line 52: Should this be exported? The `Progress` struct is part of the public API via `Status()`. Although I suppose we already have private members via `Inflights`.

tracker/progress.go
erikgrinaker: I agree that this seems a bit fragile. It would be better for the follower to communicate its commit index. It isn't clear to me that there is enough of a marginal benefit here to justify an intermediate solution prior to a protocol change.

Is this change a hard requirement for the flow control work, or simply an opportunistic improvement?

pav-kv: > Is this change a hard requirement for the flow control work, or simply an opportunistic improvement?

It is a requirement. We need to get rid of `sendIfEmpty` flag in `maybeSendAppend`, so that all the information necessary to send a message is consolidated in `Progress` struct. The decision-making for sending an empty `MsgApp` is currently not entirely encoded in `Progress`, and commit index bump is one of the cases we should "migrate" to `Progress`.

pav-kv: I can move forward with the follower->leader commit index protocol change, since it seems like a common sentiment here. This should be a small backwards-compatible PR.

tracker/progress.go
pav-kv: I'm biased towards unexporting fields by default, to avoid the risk of introducing unintended API expansion.

pav-kv: More generally, I think `Progress` needs to be split into internal and external API. For instance, all the state modifying methods are "internal" and don't need to be exported, as well as a few fields here.

erikgrinaker: Yes, but mixing private fields makes the struct a bit less convenient for external callers, e.g. it's not possible to do struct equality comparisons -- that's already not possible here though.

Being more deliberate about internal/external API here makes sense, e.g. by using a separate `ProgressStatus` struct for the public data returned by `Status()`.

tracker/progress.go
pav-kv: I added the follower commit index tracking in #132 and rebased this PR on top. However, I don't see much value in doing both things now. The `pendingCommit` index is still necessary, and remained in its original form.

pav-kv: I think we should revert this PR to the original form that was approved last week.

pav-kv: There is one case in which tracking both `Commit` and `pendingCommit` can help.

When `Commit == log.lastIndex()`, we don't need to keep sending the empty `MsgApp` periodically. However, we can only take advantage of this case in PR #134.

tracker/progress.go
pav-kv: Filed #175 to this extent.

tracker/progress.go
pav-kv: @erikgrinaker Are you happy to leave this PR as is? My assessment is that we need both parts (1) and (2), but (2) [this PR] is needed immediately. Also, (2) won't change after we add (1) - we still need to cap `pendingCommit` at `Next-1` etc.

I can maybe add some tests to check the invariants, but I already audited all the code that modifies `Next`, and made sure `pendingCommit < Next` invariant holds.

erikgrinaker: >The pendingCommit index is still necessary, and remained in its original form. Only pendingCommit is needed to decide on the next empty MsgApp message, the Commit index of the follower is not used.

Why? I'm not seeing `pendingCommit` in #132. It seems like tracking the actual acknowledged commit index would remove some assumptions, and also allow eagerly retransmitting the commit index if the previous message may have been lost (which isn't the case with `pendingCommit`).

As I understand it, the main objective here is to make `maybeSendAppend` only depend on the progress state. Isn't a follower `Commit` field sufficient for this?

pav-kv: Example why only `Commit` is not sufficient. Consider the case:
- leader log is at index 100, and commit index is 80
- follower's `Commit` is 50, and `Match` is 50

We send a `MsgApp` with entries [50,100] and commit 80. Now we need to pause sending any more messages. We will be sending periodic empty "probe" messages instead (for 2 reasons: `Match < raftLog.lastIndex` and `Commit < raftLog.committed`).

Say, the leader's commit index was updated to 100. We need to send it immediately, instead of waiting for the next "probing" attempt, because this index has a chance of applying (all entries up to 100 are in-flight too). To distinguish this case, we need the `pendingCommit < raftLog.committed` check (80 < 100). We can't use `Commit < raftLog.committed` check because we will end up spinning.

erikgrinaker (approved):

tracker/progress.go
pav-kv: > it would be nice if pendingCommit contained the actual commit index we sent (not the truncated one)

I'll consider it in #132. Not sure we should do this, because sending a commit index >= `Next` is unlikely to be applied. The leader's assumption is that the follower has all entries up to `Match`, and potentially soon will have the in-flight entries `(Match, Next)`. So if the leader's commit index is >= `Next`, it doesn't make much sense to send it more than once.

erikgrinaker: > sending a commit index >= Next is unlikely to be applied

Why do we even send it in the first place then?

tracker/progress.go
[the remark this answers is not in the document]
pav-kv: Done.

tracker/progress.go
```diff
@@ -128,27 +142,31 @@ func (pr *Progress) BecomeProbe() {
...
 	pr.Next = pr.Match + 1
+	pr.pendingCommit = min(pr.pendingCommit, pr.Next-1)
 }
 
 // BecomeSnapshot moves the Progress to StateSnapshot with the specified pending
 // snapshot index.
 func (pr *Progress) BecomeSnapshot(snapshoti uint64) {
 	pr.ResetState(StateSnapshot)
 	pr.PendingSnapshot = snapshoti
+	pr.Next = snapshoti + 1
```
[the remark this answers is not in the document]
pav-kv: Done (explanation is in the commit message). Basically, `Next` reflects the in-flight state. When we send a snapshot, this is equivalent to sending all the entries up to the snapshot index, so we set `Next = PendingSnapshot + 1`. This also allows deprecating `PendingSnapshot` in the future.

tracker/progress.go
erikgrinaker: Yes, but why don't we truncate it to Next?

pav-kv:
@ahrtr All actionable comments in this PR are resolved, and the open discussion motivates a second PR #132 after this one. This PR is ready for merge.

tracker/progress.go
pav-kv: Because there is a chance that the follower is more up-to-date than `Next`. In this case, it will apply a higher commit index, and reply with a `MsgAppResp` that resets the leader->follower flow to a more up-to-date state. This is to address the "less likely", but still possible scenario.

erikgrinaker: Seems like we shouldn't hide that fact in the field in that case, but 🤷

pav-kv: You mean we can set `sentCommit` to the actual commit index we sent? We can do that, yes. Then the condition for sending eagerly would transform into:

```golang
raftLog.committed > pr.sentCommit && pr.sentCommit < pr.Next-1
```

erikgrinaker: Yep. That seems clearer to me, but no strong opinion.

pav-kv: Updated the second commit. Now we don't need the invariant. Plus, this further reduced unnecessary sends (see `testdata` files diffs).

pav-kv: @erikgrinaker considering this resolved for now. Please comment on the updated PR if you have further suggestions.

tracker/progress.go
```diff
@@ -35,8 +35,16 @@ type Progress struct {
 	// entries with indices in (Match, Next) interval are already in flight.
 	//
 	// Invariant: 0 <= Match < Next.
+	// NB: it follows that Next >= 1.
+	//
+	// In StateSnapshot, Next == PendingSnapshot + 1.
 	Next uint64
 
+	// sentCommit is the highest commit index in flight to the follower.
```
ahrtr, line 43:
```suggestion
	// sentCommit is the highest commit index in flight to the follower.
	// It may be regressed to `Next -1` in some cases, e.g. when converting
	// to `StateProbe` or when receiving a rejection from a follower. 
```

ahrtr:
Please also rebase this PR to resolve the workflow failure.

tracker/progress.go
pav-kv: Reworded this a bit.

pav-kv:
@ahrtr done

ahrtr (approved):

PR #172, opened 2024-02-27
tracker,quorum: use native Sort

pav-kv:
The insertion sort is no longer needed, since Go 1.21 [introduced](https://tip.golang.org/doc/go1.21#slices) a generic `slices.Sort` algorithm which achieves the same effect: it doesn't incur allocations, and [uses](https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/slices/zsortordered.go;l=64-77) insertion sort for slices up to 12 elements.

### DIFF
--- a/quorum/majority.go
+++ b/quorum/majority.go
@@ -17,6 +17,7 @@ package quorum
 import (
 	"fmt"
 	"math"
+	"slices"
 	"sort"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


ahrtr:
Note: This PR will require the applications, which depend on raft, to use a go version >= 1.21. It shouldn't be a problem, because go1.20 is already out of support. But we should document the requirement in readme.

pav-kv:
`go.mod` already states 1.21, plus some other 1.21+ things are already used (e.g. #135). Seems like this PR isn't the best place to update documentation.

ahrtr:
> `go.mod` already states 1.21

It doesn't technically prevent applications, which depend on raft, from using a old go version.

> some other 1.21+ things are already used (e.g. #135). Seems like this PR isn't the best place to update documentation.

OK. We can take care of the document separately.

ahrtr (approved):

PR #160, opened 2024-02-07
Migrate to go1.22

dusk125:
Part of https://github.com/etcd-io/etcd/issues/17348

PR #178, opened 2024-03-06
dependency: update go version to 1.21.8

ivanvc:
Upgrades go version to 1.21.8.

Part of etcd-io/etcd#17533


ahrtr:
Thanks @ivanvc , but we have already bumped go version to 1.22.1
https://github.com/etcd-io/raft/pull/160

PR #179, opened 2024-03-08
Test: Replace t.error/fatal with assert/request in [node_test.go]

MrDXY:
Refactor: Replacing t.Error/t.Fatal with assert/require

This commit tries to eliminates boilerplate code in the testing suite by replacing occurrences
of t.Error and t.Fatal with assertions from the assert and require packages.

Changes Made:

- Replaced instances of t.Error with assert
- Replaced instances of t.Fatal with require

Affected Files:
- [node_test.go]

Part of: #146

### DIFF
--- a/node_test.go
+++ b/node_test.go
@@ -15,15 +15,14 @@
 package raft
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"math"
-	"reflect"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"go.etcd.io/raft/v3/raftpb"
@@ -111,9 +110,7 @@ func TestNodeStepUnblock(t *testing.T) {
 		tt.unblock()
 		select {
 		case err := <-errc:
-			if err != tt.werr {
-				t.Errorf("#%d: err = %v, want %v", i, err, tt.werr)
-			}
+			assert.Equal(t, tt.werr, err, "#%d", i)
 			// clean up side-effect
 			if ctx.Err() != nil {
 				ctx = context.TODO()
@@ -146,9 +143,7 @@ func TestNodePropose(t *testing.T) {
 	n := newNode(rn)
 	r := rn.raft
 	go n.run()
-	if err := n.Campaign(context.TODO()); err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, n.Campaign(context.TODO()))
 	for {
 		rd := <-n.Ready()
 		s.Append(rd.Entries)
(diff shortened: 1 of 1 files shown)


node_test.go
```diff
@@ -580,16 +522,12 @@ func TestNodeStart(t *testing.T) {
 
 	{
 		rd := <-n.Ready()
-		if !reflect.DeepEqual(rd, wants[0]) {
-			t.Fatalf("#1: rd = %+v,\n             w   %+v", rd, wants[0])
-		}
+		require.Equal(t, wants[0], rd, "#1")
```
pav-kv, line 525: The `#1` is unnecessary (here and below), just remove it?

If this assertion fails, the error message will contain the code line. The lines that you marked `#1`, `#2`, etc, are all different, so we'll be able to distinguish which of them failed.

Putting extra info into the message is only necessary in case the line can be called repeatedly.

node_test.go
```diff
@@ -856,9 +778,7 @@ func TestAppendPagination(t *testing.T) {
 			for _, e := range m.Entries {
 				size += len(e.Data)
 			}
-			if size > maxSizePerMsg {
-				t.Errorf("sent MsgApp that is too large: %d bytes", size)
-			}
+			assert.LessOrEqual(t, size, maxSizePerMsg, "sent MsgApp that is too large: %d bytes", size)
```
pav-kv, line 781:
```suggestion
			assert.LessOrEqual(t, size, maxSizePerMsg, "sent MsgApp that is too large")
```

An error message will already print the `size` and `maxSizePerMsg`, so no need to add it second time.

pav-kv (approved):
LGTM % nits

node_test.go
```diff
@@ -1142,18 +1046,15 @@ func TestNodeCommitPaginationAfterRestart(t *testing.T) {
...
 	defer n.Stop()
 
 	rd := readyWithTimeout(&n)
-	if !IsEmptyHardState(rd.HardState) && rd.HardState.Commit < persistedHardState.Commit {
-		t.Errorf("HardState regressed: Commit %d -> %d\nCommitting:\n%+v",
-			persistedHardState.Commit, rd.HardState.Commit,
-			DescribeEntries(rd.CommittedEntries, func(data []byte) string { return fmt.Sprintf("%q", data) }),
-		)
-	}
+	assert.False(t, !IsEmptyHardState(rd.HardState) && rd.HardState.Commit < persistedHardState.Commit,
```
ahrtr, line 1056: nit
```suggestion
	assert.True(t, IsEmptyHardState(rd.HardState) && rd.HardState.Commit < persistedHardState.Commit,
```

pav-kv: `IsEmptyHardState(rd.HardState) || rd.HardState.Commit >= persistedHardState.Commit`?

ahrtr: Good catch. I did not see the `&&`.

ahrtr: Overall not a big deal, let's keep it as it's for now. Please feel free to update it in a followup PR if you want.

ahrtr (approved):
lgtm

Thanks

PR #182, opened 2024-03-11
Test: Replace t.error/fatal with assert/request in [raft_paper_test.go]

MrDXY:
Refactor: Replacing t.Error/t.Fatal with assert/require

This commit tries to eliminate boilerplate code in the testing suite by replacing occurrences
of t.Error and t.Fatal with assertions from the assert and require packages.

Changes Made:

- Replaced instances of t.Error with assert
- Replaced instances of t.Fatal with require

Affected Files:
- [raft_paper_test.go]

Part of: #146

### DIFF
--- a/raft_paper_test.go
+++ b/raft_paper_test.go
@@ -28,9 +28,11 @@ package raft
 
 import (
 	"fmt"
-	"reflect"
 	"sort"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	pb "go.etcd.io/raft/v3/raftpb"
 )
@@ -64,12 +66,8 @@ func testUpdateTermFromMessage(t *testing.T, state StateType) {
 
 	r.Step(pb.Message{Type: pb.MsgApp, Term: 2})
 
-	if r.Term != 2 {
-		t.Errorf("term = %d, want %d", r.Term, 2)
-	}
-	if r.state != StateFollower {
-		t.Errorf("state = %v, want %v", r.state, StateFollower)
-	}
+	assert.Equal(t, uint64(2), r.Term)
+	assert.Equal(t, StateFollower, r.state)
 }
 
 // TestRejectStaleTermMessage tests that if a server receives a request with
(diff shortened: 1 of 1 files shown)


raft_paper_test.go
```diff
@@ -126,9 +120,7 @@ func TestLeaderBcastBeat(t *testing.T) {
 		{From: 1, To: 2, Term: 1, Type: pb.MsgHeartbeat},
 		{From: 1, To: 3, Term: 1, Type: pb.MsgHeartbeat},
 	}
-	if !reflect.DeepEqual(msgs, wmsgs) {
-		t.Errorf("msgs = %v, want %v", msgs, wmsgs)
-	}
+	assert.Equal(t, wmsgs, msgs)
```
pav-kv, line 123: could inline `wmsgs` as well:

```golang
assert.Equal(t, []pb.Message{
	{From: 1, To: 2, Term: 1, Type: pb.MsgHeartbeat},
	{From: 1, To: 3, Term: 1, Type: pb.MsgHeartbeat},
}, msgs)
```

Same in a few other places.

raft_paper_test.go
```diff
@@ -164,24 +156,17 @@ func testNonleaderStartElection(t *testing.T, state StateType) {
...
-	}
-	if !r.trk.Votes[r.id] {
-		t.Errorf("vote for self = false, want true")
-	}
+	assert.Equal(t, uint64(2), r.Term)
+	assert.Equal(t, StateCandidate, r.state)
+	assert.True(t, r.trk.Votes[r.id])
+
 	msgs := r.readMessages()
 	sort.Sort(messageSlice(msgs))
```
pav-kv, line 164: For another PR: consider removing `messageSlice`, and using `slices.SortFunc` with some stable comparator.

pav-kv (approved):
LGTM % some suggestions

raft_paper_test.go
MrDXY: Would this be what you had in mind?
```
slices.SortStableFunc(data, orderMessageAsc)
func orderMessageAsc(a pb.Message, b pb.Message) int {
	strA := fmt.Sprint(a)
	strB := fmt.Sprint(b)
	if strA > strB {
		return 1
	} else if strA < strB {
		return -1
	} else {
		return 0
	}
}
```

ahrtr (approved):
lgtm

Thanks

PR #204, opened 2024-06-13
Trace validation test

joshuazh-x:
Following up on previous discussion with @serathius and @ahrtr, I make some improvements to the Raft trace validation test:

* **Refactored Test Case**: Combine the random trace generation and trace validation script into a single test case for better readability and maintainability.
* **Simplified Execution**: The test can now be run with a simple `make trace-validation` command.
* **Enhanced Customization**: The test supports customization of the working directory, number of parallel jobs, and the amount of traces to be generated (see README file).
* **Other improvements**: An additional bootstrap trace has been included to simplify the TLA+ specification.

These changes aim to streamline the trace validation process and make it more efficient to automate testing.

Please let me know if you have any questions or feedback.


serathius:
Please fix the CI failures to ensure timely review.

joshuazh-x:
cc @serathius @ahrtr , here is the trace validation test with some additional improvements in TLA+ spec. The idea here is to set up a sample application based on raft, running with various faults injected. The produced traces will then be collected and validated by TLA+ trace validation. Just run `make trace-validation` in raft root folder.

zeu5:
A few changes to the TLA spec are unclear for me. I will summarise my understanding and list the concerns below. 

## Summary

Two new variables are added `applied` and `durableState.applied`. The two variables represent the state in memory and that persisted to the log. `durableState.applied` is updated on `PersistState` action and `applied` is updated on `Advance`. The rest of the changes in TLA revolve around ensuring sound updates to these two variables. 

In the code, a new trace event `Advance` is introduced to ensure updates to `applied` variable in TLA.

## Clarifications

1. Why do we need both the variables? In the code, the only `applied` state tracked is in the log. 
2. Whenever ready is requested, the `log.applied` is updated. This is inconsistent with the TLA model which requires that `applied` is updated only after the corresponding `Ready` is advanced. Is there a violation found corresponding to this inconsistency?
3. I do not see an event for `PersistState` in the trace logger. So the corresponding action is never executed?

## Minor comments

This is not related to the changes in this PR but unclear to me - when a node is stopped (as a failure introduced `makeNodesDown`) it doesn't update any variables and is stopped. In the TLA model however, this is never tracked and updates are received to the node. So the state in the code and in the model will be out of sync? Shouldn't the model track active nodes?

joshuazh-x:
Thanks @zeu5 for the summary. 

Please allow me to clarify some of my thoughts when I made this change. Please correct me if I made any mistake here.

> Why do we need both the variables? In the code, the only `applied` state tracked is in the log.
> Whenever ready is requested, the `log.applied` is updated. This is inconsistent with the TLA model which requires that `applied` is updated only after the corresponding `Ready` is advanced. Is there a violation found corresponding to this inconsistency?

To my understanding, `log.applied` is updated upon receiving `MsgStorageApplyResp`, indicating successful applying to the state machine. `log.applying` ( I assume you meant this) is updated upon ready and serves as an optimization to allow issuing new ready without waiting for complete processing of the previous one (relevant with async storage writes). Since I prioritize sync storage write in this spec so `log.applied` is used here for determining entries in ready to simplify the spec, Adjustments will be made when the spec covers async storage writes.

> I do not see an event for `PersistState` in the trace logger. So the corresponding action is never executed?

`PersistState` isn't traced as it's external to the Raft module, but it executes as part of `Next`. Initially, I aimed for non-determinism (allowing `PersistState` and `SendMessags` anywhere in the ready phase), but for TLC efficiency, I eventually choose a constrained behavior enforces sequential execution of `PersistState` and `SendMessages` after Ready. So whenever a traced `Ready` is processed, the behavior `Ready->PersistState->Ready` will be enforced.

> Shouldn't the model track active nodes?

The TLA+ spec focuses on state transitions. A non-functional node implies no state changes or message exchange, effectively modeled by behaviors excluding actions related to that node. The state space exploration will include behaviors representing such node downtime scenarios.

tla/validate-model.sh
```diff
@@ -22,16 +22,23 @@ function validate {
     local config=${2}
     local tooldir=${3}
     local statedir=${4}
+    local sim=${5:-false}
 
     set -o pipefail
-    java -XX:+UseParallelGC -cp ${tooldir}/tla2tools.jar:${tooldir}/CommunityModules-deps.jar tlc2.TLC -config "${config}" "${spec}" -lncheck final -metadir "${statedir}" -fpmem 0.9
+    if [ "$sim" ]; then
```
zeu5, line 28: The simulate flag is added to the else part. It should be here no?

tla/etcdraft.tla
```diff
@@ -103,7 +103,9 @@ VARIABLE
 VARIABLE 
     \* @type: Int -> Int;
     commitIndex
-logVars == <<log, commitIndex>>
+VARIABLE  
+    applied
```
zeu5, line 107: Missing type annotation

tla/etcdraft.tla
```diff
@@ -133,18 +135,21 @@ leaderVars == <<matchIndex, pendingConfChangeIndex>>
...
-    reconfigCount
+    appliedConfChange
 
-configVars == <<config, reconfigCount>>
+configVars == <<config, appliedConfChange>>
 
 VARIABLE 
     durableState
+VARIABLE  
+    ready
```
zeu5, line 145: Missing type annotation. Could you also add an explanation here?

zeu5 (commented):
Thank you for the clarification @joshuazh-x I have requested some additional minor changes.

I suppose it would also be good to include as part of the merge workflow some model checking on the updated TLA models to ensure the correctness of the updates. @serathius please consider this.

tla/validate-model.sh
zeu5: Also, it should be "$sim" = true, otherwise it is always true.

tla/etcdraft.tla
lemmy: TLC ignores these `@type: ...` type annotations, as they are only relevant to the Apalache model checker, which is not being used here.  As a matter of fact, they are not even part of the TLA+ language.  These annotations remain from the original upstream spec that was used as a basis for this spec.  I suggest removing all annotations.

tla/MCetcdraft.cfg
```diff
@@ -65,5 +63,4 @@ INVARIANTS
     LogMatchingInv
     QuorumLogInv
     MoreUpToDateCorrectInv
-    LeaderCompletenessInv
-    CommittedIsDurableInv
```
lemmy, line 69: Did you deliberately remove `CommittedIsDurableInv`?

tla/etcdraft.tla
```diff
@@ -248,32 +241,71 @@ InitLeaderVars == /\ matchIndex = [i \in Server |-> [j \in Server |-> 0]]
...
+    "Advance",
+    "Restart",
+    "DuplicateMessage",
+    "DropMessage",
+    "AddNewServer",
+    "AddLearner",
+    "DeleteServer"
+}
+
+IsEnabled(action, args) == action \in SpecActions
```
lemmy, line 297: `SpecActions`, `IsEnabled`, `PostAction` are orthogonal to how the raft algorithm works.  Their purpose is to constrain model checking with TLC.  Including them in the `etcdraft` specification makes the spec less readable.  For example, it would complicate Apalache-based model checking or proving the specification with TLAPS, should either be attempted in the future. Therefore, I strongly recommend limiting the constraints for model checking to the Trace* and MC* specifications.

tla/etcdraft.tla
```diff
@@ -686,43 +796,49 @@ NextAppendEntriesResponse == \E m \in DOMAIN messages : m.mtype = AppendEntriesR
...
 
 \* Defines how the variables may transition.
-NextAsync == 
+Next == 
     \/ \E i,j \in Server : RequestVote(i, j)
     \/ \E i \in Server : BecomeLeader(i)
     \/ \E i \in Server: ClientRequest(i, 0)
     \/ \E i \in Server : AdvanceCommitIndex(i)
-    \/ \E i,j \in Server : \E b,e \in matchIndex[i][j]+1..Len(log[i])+1 : AppendEntries(i, j, <<b,e>>)
+    \/ \E i,j \in Server :
```
lemmy, line 822: Typically, `Next` in large specifications only quantifies over constants to clearly indicate which variables are changed by sub-actions like `AppendEntries`.  Therefore, it would be better to move the quantification over `matchIndex` and `log` out of `Next` and into the definition of `AppendEntries`. This change will also have the side effect of making TLC display the names of sub-actions, such as `AppendEntries`, in counterexamples.

tla/etcdraft2.tla
```diff
@@ -0,0 +1,1035 @@
+--------------------------------- MODULE etcdraft2 ---------------------------------
```
lemmy, line 1: Why is there a second raft spec?

tla/validate-model.sh
joshuazh-x: You're correct. simulate flag shall be added when $sim is true. Fix it in the new commit.

tla/etcdraft.tla
joshuazh-x: Agree with @lemmy , I remove all type annotations in the new commit.

tla/etcdraft.tla
joshuazh-x: I tried to add more comments to new states and actions. Please let me know if I need to put more details here.

tla/MCetcdraft.cfg
joshuazh-x: Yes, the invariable I added previously is invalid so I remove it.

tla/etcdraft.tla
joshuazh-x: Trace* and MC* specs utilize the same mechanism of constraint in model checking. That's why I put these operators in their common base spec. In such a way we only need to focus on implementing custom IsEnabled and PostAction for different purposes. Or should I make a middle layer spec Constraint*?

tla/etcdraft2.tla
joshuazh-x: I added it by mistake. Remove it in the new commit.

tla/etcdraft.tla
joshuazh-x: I'll fix it in the new commit.

tla/etcdraft.tla
joshuazh-x: Decouple the SpecAction definitions into a new spec etcdraft_control.tla. MC* and Trace* specs extends etcdraft_control to implement their own IsEnabled and PostAction operators.

joshuazh-x:
I'll keep each fix in a separated commit during the review. If everything goes well in the end, I'll combine them into one before final merge.

tla/etcdraft.tla
joshuazh-x: Fixed in the new commit.

tla/etcdraft.tla
zeu5: Looks good to me!

lemmy (approved):

ahrtr:
Please try not to submit huge PR. Please try to breakdown this PR as much as you can,
- I see you removed the huge file `tla/example.ndjson`,  pls do it in a separate PR and ensure the readme.md is also updated accordingly.
- get the TLA+ changes in a separate PR, and leave it to TLA+ experts to review it.
- get the test included in a separate PR. I think that it makes more sense to integrate the trace test with etcd's robustness test. We can discuss that after above two items are resolved.

PR #267, opened 2025-02-15
perf: optimize MemoryStorage.Compact method by improving memory allocation

1911860538:
Avoiding the use of `append`, and directly copying remaining entries to a new slice for better performance.

### DIFF
--- a/storage.go
+++ b/storage.go
@@ -263,10 +263,9 @@ func (ms *MemoryStorage) Compact(compactIndex uint64) error {
 	// NB: allocate a new slice instead of reusing the old ms.ents. Entries in
 	// ms.ents are immutable, and can be referenced from outside MemoryStorage
 	// through slices returned by ms.Entries().
-	ents := make([]pb.Entry, 1, uint64(len(ms.ents))-i)
-	ents[0].Index = ms.ents[i].Index
-	ents[0].Term = ms.ents[i].Term
-	ents = append(ents, ms.ents[i+1:]...)
+	remainingCount := len(ms.ents) - int(i)
+	ents := make([]pb.Entry, remainingCount)
+	copy(ents, ms.ents[i:])
 	ms.ents = ents
 	return nil
 }
--- a/storage.go
+++ b/storage.go
@@ -263,7 +263,7 @@ func (ms *MemoryStorage) Compact(compactIndex uint64) error {
 	// NB: allocate a new slice instead of reusing the old ms.ents. Entries in
 	// ms.ents are immutable, and can be referenced from outside MemoryStorage
 	// through slices returned by ms.Entries().
-	remainingCount := len(ms.ents) - int(i)
+	remainingCount := uint64(len(ms.ents)) - i
 	ents := make([]pb.Entry, remainingCount)
 	copy(ents, ms.ents[i:])
 	ms.ents = ents


1911860538:
> /ok-to-test

My code update appears to be leading to unexpected issues because the original logic only copied the Term and Index of the i-th element, whereas after my modification, it now copies all fields of the i-th element.  
Perhaps the following code modification is more suitable? 
```go
ents := make([]pb.Entry, uint64(len(ms.ents)) - i)
ents[0].Index = ms.ents[i].Index
ents[0].Term = ms.ents[i].Term
copy(ents[1:], ms.ents[i:])
```

1911860538:
> > /ok-to-test
> 
> My code update appears to be leading to unexpected issues because the original logic only copied the Term and Index of the i-th element, whereas after my modification, it now copies all fields of the i-th element. Perhaps the following code modification is more appropriate?
> 
> ```go
> ents := make([]pb.Entry, uint64(len(ms.ents)) - i)
> // Only copy Index and Term of the compact index entry (first entry).
> // These fields are sufficient for log matching in the Raft algorithm,
> // while Type and Data are not needed since the entry's command has
> // already been applied to the state machine.
> ents[0].Index = ms.ents[i].Index
> ents[0].Term = ms.ents[i].Term
> copy(ents[1:], ms.ents[i+1:])
> ```

The Benchmark results of my local code indicate that this code modification results in a slight performance boost, whereas the previous code changes showed no performance improvement.  
Therefore, I am going to close this pull request. 
```go
package entry_compact

import (
	"bytes"
	"testing"
)

type Entry struct {
	Term  uint64
	Index uint64
	Type  int32
	Data  []byte
}

const (
	compactIndex = 100
	entriesLen   = 1000
)

func getEntries(length int) []Entry {
	src := make([]Entry, length)
	for i := range length {
		src[i] = Entry{
			Term:  uint64(i),
			Index: uint64(i),
			Type:  int32(i),
			Data:  []byte{'0'},
		}
	}
	return src
}

func copy1(src []Entry, index int) []Entry {
	dst := make([]Entry, 1, len(src)-index)
	dst[0].Term = src[0].Term
	dst[0].Index = src[0].Index
	dst = append(dst, src[index+1:]...)
	return dst
}

func copy2(src []Entry, index int) []Entry {
	dst := make([]Entry, len(src)-index)
	dst[0].Term = src[0].Term
	dst[0].Index = src[0].Index
	copy(dst[1:], src[index+1:])
	return dst
}

func copy3(src []Entry, index int) []Entry {
	dst := make([]Entry, len(src)-index)
	copy(dst, src[index:])
	return dst
}

func BenchmarkCopy1(b *testing.B) {
	src := getEntries(entriesLen)
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		copy1(src, compactIndex)
	}
}

func BenchmarkCopy2(b *testing.B) {
	src := getEntries(entriesLen)
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		copy2(src, compactIndex)
	}
}

func BenchmarkCopy3(b *testing.B) {
	src := getEntries(entriesLen)
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		copy3(src, compactIndex)
	}
}

func TestSame(t *testing.T) {
	src := getEntries(entriesLen)
	c1 := copy1(src, compactIndex)
	c2 := copy2(src, compactIndex)
	if len(c1) != len(c2) {
		t.Fatal("copied slices are not the same length")
	}
	for i := range c1 {
		c1i := c1[i]
		c2i := c2[i]
		if c1i.Term != c2i.Term || c1i.Index != c2i.Index || c1i.Type != c2i.Type || bytes.Compare(c1i.Data, c2i.Data) != 0 {
			t.Fatal("copied items are not same")
		}
	}
}
```

Test output
```text
=== RUN   TestSame
--- PASS: TestSame (0.00s)
PASS

Process finished with the exit code 0
```

Benchmark output
```text
goos: darwin
goarch: arm64
pkg: test_code/entry_compact
cpu: Apple M3 Pro
BenchmarkCopy1
BenchmarkCopy1-11    	  436888	      2679 ns/op	   49152 B/op	       1 allocs/op
BenchmarkCopy2
BenchmarkCopy2-11    	  452683	      2652 ns/op	   49152 B/op	       1 allocs/op
BenchmarkCopy3
BenchmarkCopy3-11    	  446326	      2681 ns/op	   49152 B/op	       1 allocs/op
PASS

Process finished with the exit code 0
```

PR #268, opened 2025-02-16
Bump go 1.24.0

Elbehery:
ref https://github.com/etcd-io/etcd/issues/19417


Elbehery:
the failing test is not caused by this PR 

```
Running [/home/runner/golangci-lint-1.60.1-linux-amd64/golangci-lint run] in [/home/runner/work/raft/raft] ...
  Error: logger.go:88:5: l.Output undefined (type *DefaultLogger has no field or method Output) (typecheck)
  		l.Output(calldepth, header("DEBUG", fmt.Sprint(v...)))
```

Elbehery:
so bumping the linter version will fix the issue

Elbehery:
oh its already here https://github.com/etcd-io/raft/pull/270 

but the linter is failing as well 😆

ivanvc:
So, `typecheck` is not a linter. I was checking the golangci-lint repository and found this: https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors

ivanvc:
I tested locally with golangci-lint v1.64.5, and it doesn't return any errors. If I run it with 1.63.4 it does.

@Elbehery, I believe you need to update to the latest golangci-lint version across the repositories, not to 1.63.4.

ahrtr (approved):

PR #273, opened 2025-02-21
chore: use go-version-file to read go-version from go.mod

mmorel-35:
#### Description

go-version can be read directly from .go-version or go.mod files using go-version-file in setup-go.

This PR uses .go-version to setup-go and allows the removal 'goversion' steps

### DIFF
--- a/.github/workflows/govuln.yaml
+++ b/.github/workflows/govuln.yaml
@@ -5,10 +5,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - id: goversion
-        run: echo "goversion=$(cat .go-version)" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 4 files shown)


.github/workflows/govuln.yaml
```diff
@@ -5,10 +5,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - id: goversion
-        run: echo "goversion=$(cat .go-version)" >> "$GITHUB_OUTPUT"
       - uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # v5.3.0
         with:
-          go-version: ${{ steps.goversion.outputs.goversion }}
+          go-version-file: .go-version
```
ahrtr, line 10: Same question: https://github.com/etcd-io/bbolt/pull/914#discussion_r1963205909?

PR #285, opened 2025-03-31
refact JointConfig to To improve code understandability

zwkno1:


confchange/confchange.go
```diff
@@ -304,10 +304,10 @@ func checkInvariants(cfg tracker.Config, trk tracker.ProgressMap) error {
 	}
 	// Conversely Learners and Voters doesn't intersect at all.
 	for id := range cfg.Learners {
-		if _, ok := outgoing(cfg.Voters)[id]; ok {
+		if _, ok := cfg.Voters.Outgoing[id]; ok {
 			return fmt.Errorf("%d is in Learners and Voters[1]", id)
```
serathius, line 308:
```suggestion
			return fmt.Errorf("%d is in Learners and Outgoing", id)
```

raft.go
```diff
@@ -1316,7 +1316,7 @@ func stepLeader(r *raft, m pb.Message) error {
 			}
 			if cc != nil {
 				alreadyPending := r.pendingConfIndex > r.raftLog.applied
-				alreadyJoint := len(r.trk.Config.Voters[1]) > 0
+				alreadyJoint := len(r.trk.Config.Voters.Incoming) > 0
```
serathius, line 1319: `Voters[1]` -> `Outgoing`?

quorum/joint.go
```diff
@@ -16,24 +16,31 @@ package quorum
...
+type JointConfig struct {
+	Incoming MajorityConfig
+	Outgoing MajorityConfig
+}
 
 func (c JointConfig) String() string {
-	if len(c[1]) > 0 {
-		return c[0].String() + "&&" + c[1].String()
+	if len(c.Outgoing) > 0 {
+		return c.Outgoing.String() + "&&" + c.Outgoing.String()
```
serathius, line 26:
```suggestion
		return c.Incoming.String() + "&&" + c.Outgoing.String()
```

serathius:
The refactor seems acceptable for merge as it was reasonable and trivial, and then I found errors.

confchange/confchange.go
```diff
@@ -121,7 +121,7 @@ func (c Changer) LeaveJoint() (tracker.Config, tracker.ProgressMap, error) {
 }
 
 // Simple carries out a series of configuration changes that (in aggregate)
-// mutates the incoming majority config Voters[0] by at most one. This method
+// mutates the incoming majority config Voters.Incomming by at most one. This method
```
serathius, line 124: Typos imply that you are doing a manual replacement, I'm worried that doing this manually will lead to errors that will be missed.

zwkno1: I think it's ok since this replacement is very small，and all tests passed.

PR #300, opened 2025-05-26
`MustSync` should return `true` if the HardState.Commit differs from previous value

ahrtr:
Even when are only applying committed entries (not appending WAL entries), the users also need to sync the `HardState` to disk as long as its value (including `Commit`) differs from previous value. Otherwise the users may run into a situation that the ApplyIndex > CommitIndex.

In example below, the last entry (a learner promotion operation) was actually committed, and also applied by etcd. But the HardState wasn't synced to disk (because `MustSync` returned false), so the commitIndex was still 9 instead of the correct value 10. Note that etcd reads the commitIndex from WAL file (from `HardState`),
```
$ ./etcd-dump-logs /tmp/etcd-etcd-1
Snapshot:
empty
Start dumping log entries from snapshot.
WAL metadata:
nodeID=b71f75320dc06a6c clusterID=1c45a069f3a1d796 term=2 commitIndex=9 vote=b71f75320dc06a6c
WAL entries: 10
lastIndex=10
term	     index	type	data
   1	         1	conf	method=ConfChangeAddNode id=b71f75320dc06a6c
   2	         2	norm	
   2	         3	norm	method=PUT path="/0/members/b71f75320dc06a6c/attributes" val="{\"name\":\"etcd-1\",\"clientURLs\":[\"http://127.0.0.1:2379\"]}"
   2	         4	norm	method=PUT path="/0/version" val="3.5.0"
   2	         5	conf	method=ConfChangeAddLearnerNode id=1bf18b91d6691abf
   2	         6	norm	method=PUT path="/0/members/1bf18b91d6691abf/attributes" val="{\"name\":\"etcd-2\",\"clientURLs\":[\"http://127.0.0.1:2378\"]}"
   2	         7	conf	method=ConfChangeAddNode id=1bf18b91d6691abf
   2	         8	conf	method=ConfChangeAddLearnerNode id=4b7ef604fa9ef7e1
   2	         9	norm	method=PUT path="/0/members/4b7ef604fa9ef7e1/attributes" val="{\"name\":\"etcd-3\",\"clientURLs\":[\"http://127.0.0.1:2377\"]}"
   2	        10	conf	method=ConfChangeAddNode id=4b7ef604fa9ef7e1

Entry types (Normal,ConfigChange) count is : 10
```

The applyIndex (consistent_index in etcd) was actualy already 10, it means that the index 10 has already been applied by etcd,

```
$ ./etcd-dump-db iterate-bucket /tmp/etcd-etcd-1/member/snap/db meta --decode
key="term", value=2
key="consistent_index", value=10
key="confState", value="{\"voters\":[2013543966895053503,5440055901155162081,13195394291058371180],\"auto_leave\":false}"
```

In etcd, we also have a use case that users can forcibly recreate a one-member cluster based on the data, no matter how many members the previous cluster has. In that cases, etcd will discard all uncommitted entries in this case and automatically remove all other members. Since previously the last HardState wasn't synced to disk, it causes the last actually committed entries being discarded (overwritten). 

When we forcibly recreated a one-member cluster,

```
$ etcd --name etcd-1 --data-dir /tmp/etcd-etcd-1 --force-new-cluster
```

etcd tried to automatically remove other members (commit two confChanges directly in a previous 3 member cluster, see index 10 and 11). Obviously previous entry with index 10 was overwritten.

```
$ ./etcd-dump-logs /tmp/etcd-etcd-1
Snapshot:
empty
Start dumping log entries from snapshot.
WAL metadata:
nodeID=b71f75320dc06a6c clusterID=1c45a069f3a1d796 term=3 commitIndex=13 vote=b71f75320dc06a6c
WAL entries: 13
lastIndex=13
term	     index	type	data
   1	         1	conf	method=ConfChangeAddNode id=b71f75320dc06a6c
   2	         2	norm	
   2	         3	norm	method=PUT path="/0/members/b71f75320dc06a6c/attributes" val="{\"name\":\"etcd-1\",\"clientURLs\":[\"http://127.0.0.1:2379\"]}"
   2	         4	norm	method=PUT path="/0/version" val="3.5.0"
   2	         5	conf	method=ConfChangeAddLearnerNode id=1bf18b91d6691abf
   2	         6	norm	method=PUT path="/0/members/1bf18b91d6691abf/attributes" val="{\"name\":\"etcd-2\",\"clientURLs\":[\"http://127.0.0.1:2378\"]}"
   2	         7	conf	method=ConfChangeAddNode id=1bf18b91d6691abf
   2	         8	conf	method=ConfChangeAddLearnerNode id=4b7ef604fa9ef7e1
   2	         9	norm	method=PUT path="/0/members/4b7ef604fa9ef7e1/attributes" val="{\"name\":\"etcd-3\",\"clientURLs\":[\"http://127.0.0.1:2377\"]}"
   2	        10	conf	method=ConfChangeRemoveNode id=1bf18b91d6691abf
   2	        11	conf	method=ConfChangeRemoveNode id=4b7ef604fa9ef7e1
   3	        12	norm	
   3	        13	norm	method=PUT path="/0/members/b71f75320dc06a6c/attributes" val="{\"name\":\"etcd-1\",\"clientURLs\":[\"http://localhost:2379\"]}"

Entry types (Normal,ConfigChange) count is : 13
```

The Index 10 has already been applied, so etcd won't re-apply it, so eventually one of the members were not removed at all,

```
$ ./etcd-dump-db iterate-bucket /tmp/etcd-etcd-1/member/snap/db members --decode
key="b71f75320dc06a6c", value="{\"id\":13195394291058371180,\"peerURLs\":[\"http://127.0.0.1:2380\"],\"name\":\"etcd-1\",\"clientURLs\":[\"http://localhost:2379\"]}"
key="1bf18b91d6691abf", value="{\"id\":2013543966895053503,\"peerURLs\":[\"http://127.0.0.1:2382\"],\"name\":\"etcd-2\",\"clientURLs\":[\"http://127.0.0.1:2378\"]}"

$ ./etcd-dump-db iterate-bucket /tmp/etcd-etcd-1/member/snap/db members_removed --decode
key="4b7ef604fa9ef7e1", value="removed"
```

@pav-kv @erikgrinaker @tbg @joshuazh-x

### DIFF
--- a/node_test.go
+++ b/node_test.go
@@ -504,7 +504,7 @@ func TestNodeStart(t *testing.T) {
 			HardState:        raftpb.HardState{Term: 2, Commit: 3, Vote: 1},
 			Entries:          nil,
 			CommittedEntries: []raftpb.Entry{{Term: 2, Index: 3, Data: []byte("foo")}},
-			MustSync:         false,
+			MustSync:         true,
 		},
 	}
 	storage := NewMemoryStorage()
--- a/rawnode.go
+++ b/rawnode.go
@@ -196,7 +196,7 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
 	// currentTerm
 	// votedFor
 	// log entries[]
-	return entsnum != 0 || st.Vote != prevst.Vote || st.Term != prevst.Term
+	return entsnum != 0 || st.Vote != prevst.Vote || st.Term != prevst.Term || st.Commit != prevst.Commit
 }
 
 func needStorageAppendMsg(r *raft, rd Ready) bool {
--- a/rawnode_test.go
+++ b/rawnode_test.go
@@ -574,7 +574,7 @@ func TestRawNodeStart(t *testing.T) {
 		HardState:        pb.HardState{Term: 1, Commit: 3, Vote: 1},
 		Entries:          nil, // emitted & checked in intermediate Ready cycle
 		CommittedEntries: entries,
-		MustSync:         false, // since we're only applying, not appending
+		MustSync:         true, // we also need to sync the hard state
 	}
 
 	storage := NewMemoryStorage()
@@ -646,7 +646,7 @@ func TestRawNodeStart(t *testing.T) {
 	require.True(t, rawNode.HasReady())
 	rd = rawNode.Ready()
 	require.Empty(t, rd.Entries)
-	require.False(t, rd.MustSync)
+	require.True(t, rd.MustSync)
 	rawNode.Advance(rd)
 
 	rd.SoftState, want.SoftState = nil, nil
--- a/testdata/async_storage_writes.txt
+++ b/testdata/async_storage_writes.txt
@@ -146,7 +146,7 @@ stabilize
 > 3 receiving messages
   AppendThread->3 MsgStorageAppendResp Term:1 Log:1/11
 > 1 handling Ready
-  Ready MustSync=false:
+  Ready MustSync=true:
   HardState Term:1 Vote:1 Commit:11
   CommittedEntries:
   1/11 EntryNormal ""
@@ -171,7 +171,7 @@ stabilize
   Responses:
   ApplyThread->1 MsgStorageApplyResp Term:0 Log:0/0 Entries:[1/11 EntryNormal ""]
 > 2 handling Ready
-  Ready MustSync=false:
+  Ready MustSync=true:
   HardState Term:1 Vote:1 Commit:11
   CommittedEntries:
   1/11 EntryNormal ""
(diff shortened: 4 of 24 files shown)


pav-kv:
Another way to go about this situation is to initialize `Commit = max(Commit, Applied)` upon restart. A higher applied index implies that the entry was committed.

Dropping entries above the `Commit` index unconditionally is unsafe anyway because the `Commit` index is always stale (even on the leader). It's only truly known to the global oracle.

ahrtr:
> A higher applied index implies that the entry was committed.

It may also possibly be a critical bug.

ahrtr:
Resolved on etcd side https://github.com/etcd-io/etcd/pull/20055

PR #302, opened 2025-06-11
chore: use Go usetesting

Elbehery:
This PR adds GO `usetesting`, to  be aligned with [etcd](https://github.com/etcd-io/etcd) repo. 

The directive itself will be added in a follow-up PR, not to fail the CI workflow.

cc @ahrtr @ivanvc


ahrtr (approved):
LGTM & thx

PR #362, opened 2025-12-15
makes sure that equal index => equal bar in quorum/majority.go Describe method

AHBICJ:
### Description
The MajorityConfig.Describe method is used to visualize commit indexes. The original comment stated that the same index would have the same bar (used for plotting), but the original implementation did not achieve this.

before
```
committed cfg=(1,2,3) cfgj=(1,4,5) idx=(100,100,99,99,98)
----
         idx
xxx>     100    (id=1)
>        100    (id=2)
x>        99    (id=3)
>         99    (id=4)
>         98    (id=5)
```

look at 100(id=2) and 99(id=4), I think they should have the same bar, as shown below.

```
committed cfg=(1,2,3) cfgj=(1,4,5) idx=(100,100,99,99,98)
----
         idx
xxx>     100    (id=1)
xxx>     100    (id=2)
x>        99    (id=3)
x>        99    (id=4)
>         98    (id=5)
```

### What does this RP do

Fix the code so that same idx have the same bar.

Change test data (joint_commit.txt and majority_commit.txt) to match the MajorityConfig.Describe method

### Why is this needed?

It's not required, but this PR will make the output more understandable.

I am not sure if these changes have any negative side effects, so I need help from someone who understands them better.
If there are side effects, or if I have misunderstood something due to my limited understanding and made a mistake, please close this PR.


serathius:
Without any context, why this is needed, what it solves, don't think it will get any review.

AHBICJ:
> Without any context, why this is needed, what it solves, don't think it will get any review.

Sorry, I made some edits and added some context.

ahrtr:
Just added an unit test for `Describe` in https://github.com/etcd-io/raft/pull/365

PR #370, opened 2026-01-15
Allow users to pass in a snapshot with only the ConfState initialized during bootstrap

ahrtr:
Link to https://github.com/etcd-io/etcd/pull/21138 and https://github.com/etcd-io/etcd/issues/20187

During etcd bootstrap, it will initialize raft's conf state using etcd v3store, in that case both the existing snapshot's index and the passed snapshot may have index 0. However there is a restriction in current implementation, the passed snapshot must have a bigger index than existing one. This PR relaxes this restriction.


ahrtr:
https://github.com/etcd-io/etcd/pull/21138 depends on this PR.

cc @fuweid @serathius @joshuazh-x  PTAL

serathius:
LGTM, PTAL @fuweid

PR #411, opened 2026-04-14
Bump go version to 1.26

ahrtr:
The main motivation is to use golang 1.26's build-in function `new`, to make the migration from gogo/protobuf to standard protoc-gen-go easier.

Usually we should stick to the oldest supported Go minor version (1.25 for now) for a library like raft, as it's up to users to use which go version. When we release raft v3.7, go 1.27 should have already released, so not a big problem.

cc @fuweid @liggitt @serathius


serathius (approved):
I'm ok with accelerating the Go bump as it benefits the main project.

fuweid (approved):

PR #412, opened 2026-04-14
Remove `gogoproto.nullable` from `Entry`

ahrtr:
Part of https://github.com/etcd-io/etcd/issues/14533


ahrtr:
cc @fuweid @liggitt

raftpb/raft_test.go
```diff
@@ -30,8 +30,9 @@ func TestProtoMemorySizes(t *testing.T) {
 		return no
 	}
 
-	var e Entry
-	assert.Equal(t, if64Bit(48, 32), unsafe.Sizeof(e), "Entry size check")
+	// TODO: use proto.Size to get Entry's size, or remove this test case
```
fuweid, line 33: Revisit this test in https://github.com/etcd-io/etcd/pull/12790.
We might need to know size change after migration.

raftpb/raft.proto
```diff
@@ -14,9 +14,9 @@ enum EntryType {
 }
 
 message Entry {
-	optional uint64     Term  = 2 [(gogoproto.nullable) = false]; // must be 64-bit aligned for atomic operations
-	optional uint64     Index = 3 [(gogoproto.nullable) = false]; // must be 64-bit aligned for atomic operations
-	optional EntryType  Type  = 1 [(gogoproto.nullable) = false];
+	optional uint64     Term  = 2; // must be 64-bit aligned for atomic operations
+	optional uint64     Index = 3; // must be 64-bit aligned for atomic operations
+	optional EntryType  Type  = 1;
```
fuweid, line 19:
```
goos: linux
goarch: amd64
pkg: go.etcd.io/raft/v3
cpu: AMD Ryzen 7 5800H with Radeon Graphics
           │ /tmp/883702c4880f0b44a242b81fde5560d2a1683654-main.txt │          /tmp/pr-412.txt           │
           │                         sec/op                         │   sec/op     vs base               │
OneNode-16                                              1.596µ ± 2%   1.642µ ± 1%  +2.91% (p=0.001 n=10)

           │ /tmp/883702c4880f0b44a242b81fde5560d2a1683654-main.txt │        /tmp/pr-412.txt         │
           │                          B/op                          │     B/op      vs base          │
OneNode-16                                             1.010Ki ± 2%   1.028Ki ± 4%  ~ (p=0.079 n=10)

           │ /tmp/883702c4880f0b44a242b81fde5560d2a1683654-main.txt │          /tmp/pr-412.txt           │
           │                       allocs/op                        │ allocs/op   vs base                │
OneNode-16                                               4.000 ± 0%   6.000 ± 0%  +50.00% (p=0.000 n=10)
```

log_test.go
```diff
@@ -954,7 +954,7 @@ func TestSlice(t *testing.T) {
 			g, err := l.slice(tt.lo, tt.hi, entryEncodingSize(tt.lim))
 			require.False(t, tt.lo <= offset && err != ErrCompacted)
 			require.False(t, tt.lo > offset && err != nil)
-			require.Equal(t, tt.w, g)
+			requireEntriesEqual(t, tt.w, g)
```
fuweid: It doesn't look good to me.

The `Size` has been changed. We should initialize Type to Normal as well.
If Type is nil, the `Size()` is different (see the code generated by protoc)

```diff
diff --git a/log_test.go b/log_test.go
index 4850525..faaf3db 100644
--- a/log_test.go
+++ b/log_test.go
@@ -877,7 +877,7 @@ func TestSlice(t *testing.T) {
        num := uint64(100)
        last := offset + num
        half := offset + num/2
-       halfe := pb.Entry{Index: new(half), Term: new(half)}
+       halfe := pb.Entry{Index: new(half), Term: new(half), Type: pb.EntryNormal.Enum()}

        entries := func(from, to uint64) []pb.Entry {
                return index(from).termRange(from, to)
@@ -954,7 +954,7 @@ func TestSlice(t *testing.T) {
                        g, err := l.slice(tt.lo, tt.hi, entryEncodingSize(tt.lim))
                        require.False(t, tt.lo <= offset && err != ErrCompacted)
                        require.False(t, tt.lo > offset && err != nil)
-                       requireEntriesEqual(t, tt.w, g)
+                       require.Equal(t, tt.w, g)
                })
        }
 }
@@ -1042,7 +1042,7 @@ func (i index) termRange(from, to uint64) []pb.Entry {
        index := uint64(i)
        entries := make([]pb.Entry, 0, to-from)
        for term := from; term < to; term++ {
-               entries = append(entries, pb.Entry{Term: new(term), Index: new(index)})
+               entries = append(entries, pb.Entry{Term: new(term), Index: new(index), Type: pb.EntryNormal.Enum()})
                index++
        }
        return entries
```

fuweid: Size change 


```diff
@@ -1526,9 +1532,15 @@ func (m *Entry) Size() (n int) {
 	}
 	var l int
 	_ = l
-	n += 1 + sovRaft(uint64(m.Type))
-	n += 1 + sovRaft(uint64(m.Term))
-	n += 1 + sovRaft(uint64(m.Index))
+	if m.Type != nil {
+		n += 1 + sovRaft(uint64(*m.Type))
+	}
+	if m.Term != nil {
+		n += 1 + sovRaft(uint64(*m.Term))
+	}
+	if m.Index != nil {
+		n += 1 + sovRaft(uint64(*m.Index))
+	}
 	if m.Data != nil {
 		l = len(m.Data)
 		n += 1 + l + sovRaft(uint64(l))
```

ahrtr: pls see the comment of `requireEntriesEqual`.  There is a TODO. After we migrate to the standard protoc-gen-go, I will update the test case.

raftpb/raft_test.go
ahrtr: same as https://github.com/etcd-io/raft/pull/412#discussion_r3088718494

log_test.go
ahrtr: To avoid long back and forth, see https://github.com/openkvlab/raft/blob/ef5934e51220fb51e9c06e261c93ac92ff4d744b/log_test.go#L881

raftpb/raft.proto
ahrtr: Did you run `BenchmarkOneNode`?

This might be expected minor performance downgrade.

diff_test.go
```diff
@@ -59,7 +59,7 @@ func ltoa(l *raftLog) string {
 	s += fmt.Sprintf("applied:  %d\n", l.applied)
 	s += fmt.Sprintf("applying:  %d\n", l.applying)
 	for i, e := range l.allEntries() {
-		s += fmt.Sprintf("#%d: %+v\n", i, e)
+		s += fmt.Sprintf("#%d: %+v\n", i, e.String())
```
liggitt, line 64: avoid depending on String() being stable, see https://github.com/protocolbuffers/protobuf-go/blob/master/internal/descfmt/stringer.go#L390-L399

log_test.go
liggitt: > The `Size` has been changed. We should initialize Type to Normal as well.

If all accessors are using the getter, unset is the same as normal, right?

Is the size or exact serialized bytes required to be identical between versions, or used for something like calculating a checksum/hash, or only the bytes in the `Data []byte` field?

liggitt: > If Type is nil, the Size() is different (see the code generated by protoc)

Correct. The nullable fields can be omitted from serialization, which modifies the serialized size. It will not change the in-memory deserialized value from a client holding old non-pointer types.

raftpb/raft.proto
```diff
@@ -14,9 +14,9 @@ enum EntryType {
 }
 
 message Entry {
-	optional uint64     Term  = 2 [(gogoproto.nullable) = false]; // must be 64-bit aligned for atomic operations
-	optional uint64     Index = 3 [(gogoproto.nullable) = false]; // must be 64-bit aligned for atomic operations
-	optional EntryType  Type  = 1 [(gogoproto.nullable) = false];
+	optional uint64     Term  = 2; // must be 64-bit aligned for atomic operations
```
liggitt: what is the significance of `must be 64-bit aligned for atomic operations`? do we know what atomic operations this is referring to that will need to be adjusted?

raftpb/raft_test.go
liggitt, line 33: was this testing sizeof for alignment reasons? if so, proto.Size doesn't make sense to use

liggitt: all of these Sizeof tests will fail when switching to protoc, which includes other fields in the generated go types

raftpb/raft.proto
fuweid: yes. I think when we use pointer, it could be memory allocation in heap. Original one might use stack memory so it doesn't need to do allocation

log_test.go
ahrtr: > To avoid long back and forth, see https://github.com/openkvlab/raft/blob/ef5934e51220fb51e9c06e261c93ac92ff4d744b/log_test.go#L881

If you read this, eventually I will use `proto.Size` to dynamically calculate the size of entries, and use `proto.Equal` to compare each entry.

diff_test.go
ahrtr: good call, added a TODO,

```
	for i, e := range l.allEntries() {
		// TODO: after migrating to protoc-gen-go, marsh entry using code like below,
		//  es, _ := proto.MarshalOptions{Deterministic: true}.Marshal(e)
		s += fmt.Sprintf("#%d: %+v\n", i, e.String())
	}
```

log_test.go
fuweid: If I understand the case correctly, it's using size to determine how many entries it will return. Even if we switch to official proto, protoc.Size will ignore unset field as well. You still need to change the case input. Without setting Normal type, the size is 6. With Normal type, it's 8. I understand why you use requireEntriesEqual to compare prefix entries. It can work. For me, if we have plan to update case in the follow up, we should keep case as its right now. At least I understand the change doesn't break anything. Just my two cents

raftpb/raft.proto
ahrtr: history: https://github.com/etcd-io/raft/commit/8aa94facb1cb212387df22b7aab88f1b2cfdf8f3

also etcd uses `atomic.Uint64` stores/loads the value.

```
	appliedIndex      atomic.Uint64
	committedIndex    atomic.Uint64
	term              atomic.Uint64
```

raftpb/raft_test.go
ahrtr: > was this testing sizeof for alignment reasons?

I don't think so. 

see the history: https://github.com/etcd-io/raft/commit/b46c140f5aa50740098f799d372c2156f5d31466

log_test.go
ahrtr: > If I understand the case correctly, it's using size to determine how many entries it will return

Yes.



> For me, if we have plan to update case in the follow up, we should keep case as its right now.

`proto.Size` and `proto.Equal` can only be used on objects which implement the `Message` interface. We wil have to update the case after migrating to the protoc-gen-go.

raftpb/raft.proto
liggitt: ok... if these are pointers, I don't think alignment matters any more, can drop those comments

util_test.go
```diff
@@ -149,3 +149,21 @@ func TestPayloadSizeOfEmptyEntry(t *testing.T) {
...
+// each Entry to string.
+// TODO: use proto.Equal to compare Entry
+func requireEntriesEqual(t *testing.T, expected, actual []pb.Entry, msgAndArgs ...interface{}) {
+	// TODO: after switching to protoc-gen-go, use proto.Size to get each entry's size,
+	// then we will be able to get the correct limit.
+	/*if len(expected) != len(actual) {
+		require.Fail(t, fmt.Sprintf("entry slice lengths differ: expected %d, got %d", len(expected), len(actual)), msgAndArgs...)
+		return
+	}*/
+	minLen := min(len(expected), len(actual))
```
liggitt: shouldn't this error if expected and actual are of different lengths?

util_test.go
```diff
@@ -149,3 +149,21 @@ func TestPayloadSizeOfEmptyEntry(t *testing.T) {
...
+	for i := range minLen {
+		if expected[i].String() != actual[i].String() {
```
liggitt: should we use gogo.Equal to compare now, and switch to proto.Equal to compare after switching to protoc?

util_test.go
ahrtr: It's still related to the discussion of https://github.com/etcd-io/raft/pull/412#discussion_r3089203815 and https://github.com/etcd-io/raft/pull/412#discussion_r3089241137. I will update the test case `TestSlice` and this function `requireEntriesEqual` after migrating to the standard protoc-gen-go.

Eventually the implementation of this function will be similar to below,

https://github.com/openkvlab/raft/blob/601865c9545df2725fb8dd00aa16f980651491d6/util_test.go#L175-L186

ahrtr: Note: I already added a comment above.

util_test.go
ahrtr: same as above https://github.com/etcd-io/raft/pull/412#discussion_r3091901525

raftpb/raft.proto
ahrtr: Done.

log_test.go
ahrtr: > If all accessors are using the getter, unset is the same as normal, right?

Yes, all reading are using getter;  by default all fields (when unset) are nil


> Is the size or exact serialized bytes required to be identical between versions, or used for something like calculating a checksum/hash, or only the bytes in the `Data []byte` field?

- The serialized bytes aren't required to be identical between versions, nor the size. 
- etcd dynamically calculates a checksum for the whole marshaled entry object data (not the entry.Data) right before writing it to WAL file. Each time after loading/unmarshaling the entry from WAL, etcd validates the entry against the checksum.

util_test.go
liggitt: it looks like the current version of this test ignores extra entries in one of the lists... that doesn't seem like a valid change to the test, even temporarily ... if we're getting lists of different lengths, isn't that a real problem?

util_test.go
liggitt: that doesn't seem valid to defer ... we should expect correct behavior, even in an intermediate state

util_test.go
ahrtr: It's because this is just an intermediate state... the test case `TestSlice` needs to be updated as mentioned above. 

I can't get all changes in one PR, otherwise it will be a super huge PR.

I have already finished the experiment (migration to protoc-gen-go) in https://github.com/openkvlab/raft...

util_test.go
ahrtr: > we should expect correct behavior, even in an intermediate state

Theoretically yes, but practically it's hard.

liggitt: if these don't pass gogo.Equal, then we shouldn't merge this PR

util_test.go
liggitt: if the slices are different lengths, then this is failing a test that should keep us from merging this PR

log_test.go
fuweid: I think it's not related to proto.size or proto.equal.

```go
halfe := pb.Entry{Index: new(half), Term: new(half), Type: pb.EntryNormal.Enum()}
```

If we use this format, it's just like non-pointer version (Type: 0 in non-pointer version). the size is `8`. we don't need to use `requireEntriesEqual` in this change because `require.Equal` is still work. Still think using `requireEntriesEqual` is not good move to make test case happy.

liggitt: shifting to compare proto equality seems like a reasonable (and even good) change to me, but that means ensuring the slice lengths are the same and gogo.Equal (or proto.Equal) returns true for pairwise comparisons

ahrtr: It's hard to make an intermediate PR fully correct on its own, and it would also cause lots of rework in later PRs.

I'm considering sending one huge complete PR instead (still commit by commit), if that works better.

liggitt: I'll defer to repo maintainers / reviewers

I'm still not sure I understand why changing fields to pointers was modifying the number of elements seen in this list ... that seems really concerning, like something somewhere isn't actually decoding the proto and is just slicing bytes on length boundaries or something

ahrtr: > I'm still not sure I understand why changing fields to pointers was modifying the number of elements seen in this list ... that seems really concerning

It isn't concerning. I am pretty sure It's caused by the test itself.

ahrtr: FYI. previously each entry has fixed size, no matter each of its field is populated or not.  Previously the teste always assume that the size of N entries is N * sizeof(entry). Now the size is dynamical, so the size of N entries is uncertain now; we need to dynamically calculate the size.

fuweid: Besides this case, the changes looks good to me.

* it's compatible with existing format (current pr <---> 3.6.0)

But I can't make LGTM on this change because of this test (even if my lgtm is not-binding :p) 

> Now the size is dynamical

In this case, the size is `6`. it's fixed. That's why I suggest making entry with `Type` and it will be fixed size `8`. And then we can mark this case with TODO or FIXME and refactor it in the follow-up.

>  it would also cause lots of rework in later PRs.

Let's other maintainers to chime.

util_test.go
ahrtr: removed this function.

util_test.go
ahrtr: resolved

log_test.go
ahrtr: resolved.

ahrtr:
cc. @fuweid @liggitt I believe I resolved all comment, PTAL

cc @serathius

fuweid (approved):

ahrtr:
@liggitt does this PR look good to you now?

liggitt:
> @liggitt does this PR look good to you now?

yes, lgtm

serathius (approved):

PR #414, opened 2026-04-15
Align formatting of node ID in MajorityConfig

### LINKED ISSUE #413: Configuration change logging uses inconsistent node ID formatting
Node IDs are almost always formatted as hexadecimals in Raft's logs. Except for logging configuration changes, where they are logged as decimals. This makes it harder to track what's going on with configuration changes, because I can't do hex <-> decimal conversions in my head.

Current:
```
raft2026/04/15 08:08:48 INFO: e9 switched to configuration voters=(54 233 533)
```

Desired:
```
raft2026/04/15 08:08:48 INFO: e9 switched to configuration voters=(36 e9 215)
```

This doesn't affect tests from what I grepped. All test data looking for these log lines uses node IDs between 1 and 3, where the formatting is the same.

robinkb:
Resolves #413.

### DIFF
--- a/quorum/majority.go
+++ b/quorum/majority.go
@@ -37,7 +37,7 @@ func (c MajorityConfig) String() string {
 		if i > 0 {
 			buf.WriteByte(' ')
 		}
-		fmt.Fprint(&buf, sl[i])
+		fmt.Fprintf(&buf, "%x", sl[i])
 	}
 	buf.WriteByte(')')
 	return buf.String()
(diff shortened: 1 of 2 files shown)


quorum/majority.go
```diff
@@ -37,7 +37,7 @@ func (c MajorityConfig) String() string {
 		if i > 0 {
 			buf.WriteByte(' ')
 		}
-		fmt.Fprint(&buf, sl[i])
+		fmt.Fprintf(&buf, "%x", sl[i])
```
ahrtr, line 40: You might want to update line 103 as well, to keep them consistent.

```
fmt.Fprintf(&buf, " %5d    (id=%x)\n", info[i].idx, info[i].id)
```

This is a minor behaviour change. Please also add an entry into https://github.com/etcd-io/raft/blob/main/CHANGELOG/CHANGELOG-3.7.md. thx

robinkb: Done. Thanks for the fast review!

ahrtr (approved):

PR #416, opened 2026-05-01
Remove `gogoproto.nullable` option from `SnapshotMetadata` and `Snapshot`

ahrtr:


ahrtr:
cc @fuweid @liggitt

raft.go
```diff
@@ -1867,7 +1867,7 @@ func (r *raft) restore(s pb.Snapshot) bool {
 	// config. This shouldn't ever happen (at the time of writing) but lots of
 	// code here and there assumes that r.id is in the progress tracker.
 	found := false
-	cs := s.Metadata.ConfState
```
fuweid, line 1870: I'm not sure it's likely. so maybe we should fast-return here if cs is nill.

storage.go
```diff
@@ -239,10 +243,11 @@ func (ms *MemoryStorage) CreateSnapshot(i uint64, cs *pb.ConfState, data []byte)
...
 	}
 
-	ms.snapshot.Metadata.Index = i
-	ms.snapshot.Metadata.Term = ms.ents[i-offset].GetTerm()
+	pb.EnsureSnapshot(&ms.snapshot)
+	ms.snapshot.Metadata.Index = new(i)
+	ms.snapshot.Metadata.Term = new(ms.ents[i-offset].GetTerm())
 	if cs != nil {
-		ms.snapshot.Metadata.ConfState = *cs
+		ms.snapshot.Metadata.ConfState = cs
```
fuweid: etcd is using one goroutine to handle apply logic. so it's unlikely to see one goroutine is marshaling snapshot and other one is updating the ConfState. We can consider doing shadow-copy here.

fuweid (approved):
LGTM with two comments

raft.go
ahrtr: It will never nil, see `EnsureSnapshotMetadata`

serathius (approved):

storage.go
ahrtr: updated to deep copy the provided confstate.

fuweid (approved):

PR #423, opened 2026-05-13
Remove `gogoproto.nullable` from `ConfState`

ahrtr:


ahrtr:
cc @fuweid @liggitt

raft.go
```diff
@@ -1872,6 +1872,7 @@ func (r *raft) restore(s pb.Snapshot) bool {
 	// code here and there assumes that r.id is in the progress tracker.
 	found := false
 	cs := s.GetMetadata().GetConfState()
+	cs = pb.EnsureConfState(cs)
```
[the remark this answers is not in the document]
ahrtr: > Consider normalizing the InitialState conf state as well (e.g., call pb.EnsureConfState(&cs) right after InitialState()), so external Storage implementations remain compatible.

Already added.

serathius (approved):

raftpb/confstate_test.go
```diff
@@ -47,7 +47,7 @@ func TestConfState_Equivalent(t *testing.T) {
 		// Non-equivalent learners.
 		{ConfState{Voters: []uint64{1, 2, 3, 4}}, ConfState{Voters: []uint64{2, 1, 3}}, false},
 		// Sensitive to AutoLeave flag.
-		{ConfState{AutoLeave: true}, ConfState{}, false},
```
fuweid, line 50: We should be careful to handle `ConfState{AutoLeave: new(false)}` compared with `ConfState{AutoLeave: nil}`. I think `ConfState.Equivalent` should treat those as equivalent, since both mean `AutoLeave == false`. Today it sorts the slices and then uses `reflect.DeepEqual`, so these two semantically identical states compare unequal. Maybe we should switch this to an explicit field-by-field comparison and compare AutoLeave via `GetAutoLeave()`. For example,

```go
 func (cs ConfState) Equivalent(cs2 ConfState) error {
        orig1, orig2 := cs, cs2

        sorted := func(sl []uint64) []uint64 {
                out := append([]uint64(nil), sl...)
                slices.Sort(out)
                return out
        }

        equal := slices.Equal(sorted(cs.Voters), sorted(cs2.Voters)) &&
                slices.Equal(sorted(cs.Learners), sorted(cs2.Learners)) &&
                slices.Equal(sorted(cs.VotersOutgoing), sorted(cs2.VotersOutgoing)) &&
                slices.Equal(sorted(cs.LearnersNext), sorted(cs2.LearnersNext)) &&
                cs.GetAutoLeave() == cs2.GetAutoLeave()

        if !equal {
                return fmt.Errorf("ConfStates not equivalent:\n%+#v\n%+#v", orig1, orig2)
        }
        return nil
  }
```

fuweid (approved):
LGTM

raftpb/confstate_test.go
ahrtr: pls see `EnsureConfState` in util.go. The `AutoLeave` will never be nil in production.

Updated this test to ensure it to be a non-nil value using `EnsureConfState`

fuweid: I mean that `Equivalent` is public method to user. They can invoke this without EnsureConfState.

ahrtr: Currently it's only used by test & assert in this repo, it should be OK. 

Let me add a TODO to call `EnsureConfState` inside this method.

PR #425, opened 2026-05-15
Remove `gogoproto.nullable` from `ConfChangeSingle` and `ConfChangeV2`

ahrtr:
Part of https://github.com/etcd-io/etcd/issues/14533

assisted-by: Copilot


ahrtr:
@fuweid @liggitt @serathius

raftpb/confchange.go
```diff
@@ -107,9 +105,8 @@ func (c ConfChangeV2) EnterJoint() (autoLeave bool, ok bool) {
 // This is the case if the ConfChangeV2 is zero, with the possible exception of
 // the Context field.
 func (c ConfChangeV2) LeaveJoint() bool {
-	// NB: c is already a copy.
-	c.Context = nil
-	return proto.Equal(&c, &ConfChangeV2{})
+	return c.GetTransition() == ConfChangeTransition_ConfChangeTransitionAuto &&
```
serathius, line 108: This change is unusual.

ahrtr: I intentionally made this change. It doesn't care about the field `Context`, we just need to check other two fields. We are good as long as it's an empty `ConfChangeV2`

serathius: This approach seems less future proof than comparing to empty struct. Do we have a test that would ensure that we don't break the behavior in the future?

ahrtr: added a test case TestLeaveJoint in the last commit

serathius: Thanks

serathius (approved):

fuweid (approved):

PR #435, opened 2026-05-22
Switch to standard protoc-gen-to

ahrtr:


serathius (approved):

scripts/genproto.sh
```diff
@@ -18,27 +18,25 @@ if [[ $(protoc --version | cut -f2 -d' ') != "3.20.3" ]]; then
...
 for dir in ${DIRS}; do
   pushd "${dir}"
-    protoc --gofast_out=. -I=".:${GOGOPROTO_PATH}:${RAFT_ROOT_DIR}/..:${RAFT_ROOT_DIR}" \
-      --plugin="${GOFAST_BIN}" ./**/*.proto
-
-    sed -i.bak -E 's|"raft/raftpb"|"go.etcd.io/etcd/raft/v3/raftpb"|g' ./**/*.pb.go
-    sed -i.bak -E 's|"google/protobuf"|"github.com/gogo/protobuf/protoc-gen-gogo/descriptor"|g' ./**/*.pb.go
+    protoc --go_out=. -I=".:${GOGOPROTO_PATH}:${RAFT_ROOT_DIR}/..:${RAFT_ROOT_DIR}" \
+    --plugin=protoc-gen-go="${GOGEN_BIN}" \
+    --go_opt=paths=source_relative ./**/*.proto
```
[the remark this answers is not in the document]
ahrtr: resolved

ahrtr:
`github.com/gogo/protobuf` is entirely gone now

tools/mod/go.mod
```diff
@@ -8,7 +8,6 @@ require (
 	github.com/alexkohler/nakedret v1.0.0
 	github.com/chzchzchz/goword v0.0.0-20170907005317-a9744cb52b03
 	github.com/coreos/license-bill-of-materials v0.0.0-20190913234955-13baff47494e
-	github.com/gogo/protobuf v1.3.2
```
liggitt, line 11: 🎉

go.mod
```diff
@@ -6,15 +6,13 @@ toolchain go1.26.3
 
 require (
 	github.com/cockroachdb/datadriven v1.0.2
-	github.com/gogo/protobuf v1.3.2
-	github.com/golang/protobuf v1.5.4
```
liggitt, line 10: 🎉

go.mod
```diff
@@ -6,15 +6,13 @@ toolchain go1.26.3
...
 	github.com/stretchr/testify v1.11.1
+	google.golang.org/protobuf v1.33.0
```
liggitt: should we make this match the `google.golang.org/protobuf v1.36.11` used in tools.go?

storage_test.go
```diff
@@ -65,18 +66,18 @@ func TestStorageEntries(t *testing.T) {
 	}{
 		{2, 6, math.MaxUint64, ErrCompacted, nil},
 		{3, 4, math.MaxUint64, ErrCompacted, nil},
-		{4, 5, math.MaxUint64, nil, index(4).terms(4)},
-		{4, 6, math.MaxUint64, nil, index(4).terms(4, 5)},
-		{4, 7, math.MaxUint64, nil, index(4).terms(4, 5, 6)},
+		{4, 5, math.MaxUint64, nil, ents[1:2]},
```
liggitt: I'm curious about these changes ... why are we not constructing distinct test entries any more but are reusing entries from the `ents` slice? doesn't that risk cross case pollution?

storage_test.go
liggitt: should this switch to requireEqualEntries?

liggitt (commented):
looks great, just a few questions

go.mod
ahrtr: Good catch, updated to v1.36.11.

storage_test.go
ahrtr: Actually the entries are readonly in this case. reverted the change to avoid confusion. 

Also updated to use `requireEqualEntries`

storage_test.go
ahrtr: see above comment, updated to use `requireEqualEntries`

storage_test.go
liggitt: sounds good, easier to be sure no in-place mutation is happening if we just keep the instances separate

PR #445, opened 2026-06-03
[release-3.6] bump go 1.25.11

pjsharath28:
Part of https://github.com/etcd-io/etcd/issues/21885


ivanvc:
@pjsharath28, do we need the `setup-go` version bumps? Or can we drop them?

ahrtr (approved):

pjsharath28:
> @pjsharath28, do we need the setup-go version bumps? Or can we drop them?

Sorry, it turns out the setup-go bumps were not actually required. I initially hit some workflow failures, and upgrading `setup-go` appeared to help, so I assumed it might be related. After reverting the changes, the workflows still passed.

Thanks for pointing that out.

Issue #456, opened 2026-06-22
[Spec Bug] TLA+ Raft model permits split-brain during joint/membership changes due to missing log guard


CharlieQiu2017:
Hello, while I was auditing this Raft implementation I noticed a safety bug in the current TLA+ specification of Raft, specifically in the member change logic. This issue **does not** affect the actual Go code, which already contains necessary guards to prevent it. The issue is due to the TLA+ model not accurately representing the current implementation.

In more detail: recall that the current member config change is implemented as follows:
1. Unlike the standard Raft algorithm, in the etcd implementation a reconfig entry does not take effect until it is committed.
In the TLA+ model this occurs in two steps.
First the follower receives a heartbeat message and updates its `commitIndex` (see `AppendEntriesAlreadyDone()` and `CommitTo()`).
Then the follower executes `ApplySimpleConfChange()` to update the config.
2. When any follower times out and initiates a leader campaign, the config it uses to count votes is the latest config it has applied,
which is in general not the same as the latest config it its local log. In fact, the latest applied config could lag behind the latest
config entry by several versions.
3. In the actual etcd implementation, when a follower initiates a leader campaign, it needs to check whether there exists committed but
unapplied reconfig entries in its local log. See `hup()` which calls `hasUnappliedConfChanges()` in `raft.go`.
If so, the campaign is abandoned. Since the leader enforces that there can be at most one uncommitted reconfig entry,
this effectively means the config used by the follower can lag behind the latest config by at most one version.
However, this check seems to be not reflected in the TLA+ model. Notice that neither `Timeout()` nor `RequestVote()` has any checks over reconfig entries in the local log.

This leads the following possible sequence of events in the TLA+ model:
1. Initially, the system consists of s1, s2, s3, and leader is s1.
2. The leader proposes a reconfig entry so the system now consists of s1, s2, s3, s4, and the reconfig is committed but not applied.
3. The leader proposes a second reconfig entry so the system now consists of s1, s2, s3, s4, s5, and this second reconfig is also committed but not applied.
4. s2 now applies the second reconfig entry. However s3 applies neither of the reconfig entries.
5. Both s2 and s3 now initiate a leader campaign using the latest log (which contains both reconfig entries).
Remember that the config of s2 is {s1, s2, s3, s4, s5}, and the config of s3 is {s1, s2, s3}.
[...]

PR #458, opened 2026-06-29
raft: fix TLA+ specification to match unapplied config guard

CharlieQiu2017:
As described in issue #456, the current TLA+ model of this Raft implementation is subtly broken due to a missing guard condition for leader campaigns. In the actual implementation, a follower or pre-candidate needs to check that there are no unapplied committed reconfig entries in its local log before initiating a campaign, but this check is not enforced in the model, leading to a potential safety violation. This PR fixes the issue by adding the guard condition. Running the driver spec in #456 against the updated TLA+ model shows that the safety violation is avoided.


ahrtr:
cc @joshuazh-x

PR #462, opened 2026-07-04
Remove unnecessary log shrinking

### LINKED ISSUE #71: Release in-memory log space based on entries size

MohanadKh03:
Resolves: https://github.com/etcd-io/raft/issues/71

### DIFF
--- a/CHANGELOG/CHANGELOG-3.8.md
+++ b/CHANGELOG/CHANGELOG-3.8.md
@@ -1,0 +1,13 @@
+
+<hr>
+
+## v3.8.0
+
+### Bug Fixes
+-
+
+### Features
+-
+
+### Others
+- [Remove unnecessary log shrinking](https://github.com/etcd-io/raft/pull/462)
--- a/log_unstable.go
+++ b/log_unstable.go
@@ -164,21 +164,12 @@ func (u *unstable) stableTo(id entryID) {
 }
 
 // shrinkEntriesArray discards the underlying array used by the entries slice
-// if most of it isn't being used. This avoids holding references to a bunch of
+// if it isn't being used. This avoids holding references to a bunch of
 // potentially large entries that aren't needed anymore. Simply clearing the
 // entries wouldn't be safe because clients might still be using them.
 func (u *unstable) shrinkEntriesArray() {
-	// We replace the array if we're using less than half of the space in
-	// it. This number is fairly arbitrary, chosen as an attempt to balance
-	// memory usage vs number of allocations. It could probably be improved
-	// with some focused tuning.
-	const lenMultiple = 2
 	if len(u.entries) == 0 {
 		u.entries = nil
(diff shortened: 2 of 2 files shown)


log_unstable.go
```diff
@@ -168,17 +168,12 @@ func (u *unstable) stableTo(id entryID) {
...
 // entries wouldn't be safe because clients might still be using them.
 func (u *unstable) shrinkEntriesArray() {
-	// We replace the array if we're using less than half of the space in
-	// it. This number is fairly arbitrary, chosen as an attempt to balance
-	// memory usage vs number of allocations. It could probably be improved
-	// with some focused tuning.
+	// TODO: revisit shrinking. Using len/cap wasn't useful since
+	// stableTo slices the entries. It might be better to shrink 
+	// based on the actual size of the retained entries instead.
 	const lenMultiple = 2
```
ahrtr: We don't have plan to spend too much effort to optimize this. Let's just keep it as simple as possible.
```suggestion
```

log_unstable.go
ahrtr:
```suggestion
// if it isn't being used. This avoids holding references to a bunch of
```

ahrtr:
Please also add a file CHANGELOG-3.8.md under https://github.com/etcd-io/raft/tree/main/CHANGELOG, and add an item for this change

CHANGELOG/CHANGELOG-3.8.md
```diff
@@ -0,0 +1,13 @@
...
+## v3.8.0
+
+### Bug Fixes
+-
+
+### Features
+-
+
+### Others
+- [Update unstable log shrinking](https://github.com/etcd-io/raft/pull/462)
```
ahrtr:
```suggestion
- [Remove unnecessary log shrinking](https://github.com/etcd-io/raft/pull/462)
```

ahrtr (approved):

PR #479, opened 2026-07-25
log: optimize findConflictByTerm with binary search

### LINKED ISSUE #180: optimization：add preTermLastlog index to accelerate the find hintIndex
Now etcd raft find hintindex slowly
```go
func (l *raftLog) findConflictByTerm(index uint64, term uint64) (uint64, uint64) {
	for ; index > 0; index-- {
		// If there is an error (likely ErrCompacted or ErrUnavailable), we don't
		// know whether it's a match or not, so assume a possible match and return
		// the index, with 0 term indicating an unknown term.
		if ourTerm, err := l.term(index); err != nil {
			return index, 0
		} else if ourTerm <= term {
			return index, ourTerm
		}
	}
	return 0, 0
}
```
But when PreVote and checkQuorum is enable, one electionTimeout's log may be the stale log When partition happened.This is common.And When network recover, the follower will search many logs. And When PreVote and CheckQuorum is enable,  there is only one term logs will be stale in the most cases. So I think there is a need to add preTermLastLog to accelerate this.For example:

```go
func (l *raftLog) findConflictByTerm(index uint64, term uint64) (uint64, uint64) {
	if l.preTermLastIndex > 0 {
		return l.preTermLastIndex, l.preTerm 
	}
	for ; index > 0; index-- {
		// If there is an error (likely ErrCompacted or ErrUnavailable), we don't
		// know whether it's a match or not, so assume a possible match and return
		// the index, with 0 term indicating an unknown term.
		if ourTerm, err := l.term(index); err != nil {
			return index, 0
		} else if ourTerm <= term {
			return index, ourTerm
		}
	}
	return 0, 0
}
```

Lfan-ke:
The current `findConflictByTerm` performs an O(n) linear scan backward from the
given index. Since log terms are monotonically non-decreasing, we can instead
binary search over `[firstIndex-1, index]` to find the rightmost entry whose
term does not exceed the target, reducing the worst-case complexity from O(n)
to O(log n).

This matters most during partition recovery, where a follower's conflicting
suffix may span thousands of entries.

The valid term range is `[firstIndex-1, lastIndex]`:
- `firstIndex-1` is the snapshot index, whose term is still readable.
- Entries below `firstIndex-1` are fully compacted and cause `ErrCompacted`.
- Entries above `lastIndex` cause `ErrUnavailable`.

Corner cases handled:
- `index > lastIndex`: return `(index, 0)` immediately (ErrUnavailable).
- `index < firstIndex-1`: return `(index, 0)` immediately (ErrCompacted).
- All accessible entries exceed `term`: return `(firstIndex-2, 0)` to signal
  that the matching point is in compacted territory, matching the original
  linear-scan behavior.

All existing `TestFindConflictByTerm` cases pass unchanged. Full test suite
passes with `go test ./...`.

Closes #180

### DIFF
--- a/log.go
+++ b/log.go
@@ -180,17 +180,44 @@ func (l *raftLog) findConflict(ents []*pb.Entry) uint64 {
 // an unsuccessful append to a follower, and ultimately restore the steady flow
 // of appends.
 func (l *raftLog) findConflictByTerm(index uint64, term uint64) (uint64, uint64) {
-	for ; index > 0; index-- {
-		// If there is an error (likely ErrCompacted or ErrUnavailable), we don't
-		// know whether it's a match or not, so assume a possible match and return
-		// the index, with 0 term indicating an unknown term.
-		if ourTerm, err := l.term(index); err != nil {
-			return index, 0
-		} else if ourTerm <= term {
-			return index, ourTerm
+	if index > l.lastIndex() {
+		// Beyond the end of the log; the caller must treat 0 term as unknown.
+		return index, 0
+	}
+	// lo is the snapshot index (firstIndex-1), whose term is accessible even
+	// though the entry itself is compacted. Terms are monotonically non-decreasing
+	// across [lo, lastIndex], so binary search is valid.
+	lo := l.firstIndex() - 1
+	if index < lo {
+		// Fully compacted; term is unknown.
+		return index, 0
+	}
+	// Binary search for the rightmost position in [lo, index] where term(i) <= term.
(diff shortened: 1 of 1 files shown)


Lfan-ke:
Closing this one. Took another look and the binary-search rewrite isn't correct at the snapshot boundary - the index underflows to a garbage value where the linear scan returns 0, so it would fail the existing findConflictByTerm cases once CI runs. Not worth keeping open as-is.

